PR #369, opened 2015-01-02
build_image: add ability to pick different base package

polvi:
This allows us to play with different image bases.


marineam:
LGTM but perhaps update the usage text to note that this only applies to prod images.

polvi:
done, thanks.

PR #370, opened 2015-01-03
Extra portdir support

techdragon:
Adding the ability to specify an extra portdir_overlay directory into a CoreOS build. This will be useful functionality for all the people who want to build otherwise vanilla CoreOS images, but with additional software for whatever they are trying to do. There is an associated pull request on the chromite repo. coreos/chromite#18

### DIFF
--- a/sdk_lib/make_chroot.sh
+++ b/sdk_lib/make_chroot.sh
@@ -159,7 +159,7 @@ init_setup () {
    ln -sf "${CHROOT_TRUNK_DIR}/src/third_party/portage-stable" \
      "${FLAGS_chroot}"/"${PORTAGE_STABLE_OVERLAY}"
 
-   if [[ -n "${FLAGS_extra_portdir_overlay_path }" ]]; then
+   if [[ -n "${FLAGS_extra_portdir_overlay_path}" ]]; then
      if [[ -n "${FLAGS_extra_portdir_overlay_name}" ]]; then
        ln -sf "${CHROOT_TRUNK_DIR}"/"${FLAGS_extra_portdir_overlay_path}" \
          "${FLAGS_chroot}"/usr/local/portage/"${FLAGS_extra_portdir_overlay_name}"
--- a/sdk_lib/make_chroot.sh
+++ b/sdk_lib/make_chroot.sh
@@ -47,6 +47,11 @@ DEFINE_string stage3_path "" \
   "Use the stage3 located on this path."
 DEFINE_string cache_dir "" "Directory to store caches within."
 
+DEFINE_string extra_portdir_overlay_path "" \
+  "Extra PORTDIR_OVERLAY direcory path outside of the chroot."
+DEFINE_string extra_portdir_overlay_name "" \
+  "Name of the symlink directory that will be created inside the chroot."
+
 # Parse command line flags.
 FLAGS_HELP="usage: $SCRIPT_NAME [flags]"
 FLAGS "$@" || exit 1
@@ -153,6 +158,13 @@ init_setup () {
      "${FLAGS_chroot}"/"${CHROOT_OVERLAY}"
    ln -sf "${CHROOT_TRUNK_DIR}/src/third_party/portage-stable" \
      "${FLAGS_chroot}"/"${PORTAGE_STABLE_OVERLAY}"
+
+   if [[ -n "${FLAGS_extra_portdir_overlay_path }" ]]; then
+     if [[ -n "${FLAGS_extra_portdir_overlay_name}" ]]; then
+       ln -sf "${CHROOT_TRUNK_DIR}"/"${FLAGS_extra_portdir_overlay_path}" \
+         "${FLAGS_chroot}"/usr/local/portage/"${FLAGS_extra_portdir_overlay_name}"
+     fi
+   fi
 
    # Set up sudoers.  Inside the chroot, the user can sudo without a password.
    # (Safe enough, since the only way into the chroot is to 'sudo chroot', so
@@ -317,6 +329,11 @@ else
 fi
 if [[ "${FLAGS_jobs}" -ne -1 ]]; then
   UPDATE_ARGS+=( --jobs=${FLAGS_jobs} )
+fi
+if [[ -n "${FLAGS_extra_portdir_overlay_path }" ]]; then
+  if [[ -n "${FLAGS_extra_portdir_overlay_name}" ]]; then
+    UPDATE_ARGS+=( "--extra_portdir_overlay_dir=/usr/local/portage/${FLAGS_extra_portdir_overlay_name}" )
+  fi
 fi
 enter_chroot "${CHROOT_TRUNK_DIR}/src/scripts/update_chroot" "${UPDATE_ARGS[@]}"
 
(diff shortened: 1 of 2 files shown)


sdk_lib/make_chroot.sh
```diff
@@ -47,6 +47,11 @@ DEFINE_string stage3_path "" \
   "Use the stage3 located on this path."
 DEFINE_string cache_dir "" "Directory to store caches within."
 
+DEFINE_string extra_portdir_overlay_path "" \
```
marineam: I think `local_overlay` would be a more convenient name for this.

sdk_lib/make_chroot.sh
```diff
@@ -47,6 +47,11 @@ DEFINE_string stage3_path "" \
...
+  "Extra PORTDIR_OVERLAY direcory path outside of the chroot."
+DEFINE_string extra_portdir_overlay_name "" \
```
marineam: Perhaps just use the name in the overlay's `profiles/repo_name` file and use the current directory name as a fallback. That'll simplify using this feature.

sdk_lib/make_chroot.sh
```diff
@@ -154,6 +159,13 @@ init_setup () {
    ln -sf "${CHROOT_TRUNK_DIR}/src/third_party/portage-stable" \
      "${FLAGS_chroot}"/"${PORTAGE_STABLE_OVERLAY}"
 
+   if [[ -n "${FLAGS_extra_portdir_overlay_path}" ]]; then
+     if [[ -n "${FLAGS_extra_portdir_overlay_name}" ]]; then
+       ln -sf "${CHROOT_TRUNK_DIR}"/"${FLAGS_extra_portdir_overlay_path}" \
```
marineam: How about do this in update_chroot so you can an overlay after the initial chroot creation.

marineam: Actually, or just not bother with a symlink at all. Just make sure the overlay path is somewhere in the source tree (and thus accessible in the chroot) and translate the path to the right location under /mnt/host before passing it to update_chroot.

update_chroot
```diff
@@ -58,7 +66,7 @@ sudo touch "/etc/portage/make.conf.user"
 sudo_clobber "/etc/portage/make.conf" <<EOF
 # Created by update_chroot
 PORTDIR="${PORTAGE_STABLE_OVERLAY}"
-PORTDIR_OVERLAY="${CROSSDEV_OVERLAY} ${COREOS_OVERLAY}"
+PORTDIR_OVERLAY="${CROSSDEV_OVERLAY} ${COREOS_OVERLAY} ${EXTRA_PORTDIR_OVERLAY}"
```
marineam: The PORTDIR* variables are actually deprecated and at some point I should prune references to them. The newer scheme is to add configs under repos.conf. Does it work to just write out /etc/portage/repos.conf/local.conf and not bother with adding it to the variable? A special config file would have the added benefit preserving this setting when update_chroot gets called later, likely without the new overlay option. As is this setting will get lost as soon as you run build_packages.

sdk_lib/make_chroot.sh
techdragon: I agree, but was wary of using too generic a name since I didn't want to cause possible conflicts anywhere else.

sdk_lib/make_chroot.sh
techdragon: I did contemplate something like this, just specifying the name of the overlay and then assuming that it is located in ```src/third_party``` as ```src/third_party/overlay-name``` but again, I was trying to not 'over assume' since despite building VMs for months, I haven't needed to tamper with any of this code at all before.

sdk_lib/make_chroot.sh
marineam: Generic is fine, I don't think there is anything else that this would conflict with or be confused with. There used to be other more complicated schemes for pulling in other overlays but that all broke and was pulled out some time ago.

sdk_lib/make_chroot.sh
techdragon: Just checking I follow you here, are you suggesting to set the raw path e.g. ```/mnt/host/.../overlay-name``` directly to the ```PORTDIR_OVERLAY``` variable to avoid the need for the symlink?

sdk_lib/make_chroot.sh
marineam: Well, assuming src/third_party may be a bit much but expecting a full path as an argument and then enforcing that that path is inside of the source tree should be reasonable. Otherwise you will need to do extra steps like bind mounting the path into the chroot.

update_chroot
techdragon: I didn't think about repos.conf ... that could have saved some time since the entire point is to enable this 'extra dir' for anyone looking to 'bolt on' to CoreOS, and backtracking through the build code, the PORTDIR_OVERLAY flag in make.conf seemed the place to fix.

I did think about the 'wipeout' problem and decided that anyone using these options would be smart enough to ensure they keep the command line flags in order and used the option everywhere necessary. But repos.conf does offer a much cleaner way.

sdk_lib/make_chroot.sh
techdragon: that was why i assumed it was inside the work dir, represented by src/.. chromite/.. and chroot/.. (their shared parent)

sdk_lib/make_chroot.sh
marineam: Yes, I may do the same with portage-stable and coreos-overlay at some point.

sdk_lib/make_chroot.sh
marineam: Ah, I missed that detail. Would be good to resolve the path with something like `readlink -f` and explicitly test for that assumption, otherwise it will mysteriously break for someone else down the road.

techdragon: I did try to make sure the options help text was explicit enough, but your right, testing the assumption would be better, I was just worried about adding too much logic to the bash scripts.

techdragon:
@marineam the repos.conf file offers a possible rewrite, but it does require a slightly more 'major' change than the simple addition of new functionality I have here, and I'm not sure how the team might feel about it.
By specifying a location for 'repo configuration files' we can modify the pipeline so that when it populates the repos.conf/coreos.conf it also symlinks any files matching ```whatever_we_call_the_repo_conf_dir/*/*.conf``` into ```/etc/portage/repos.conf/``` which would provide everything these two pull requests do, and probably more. 

What do you think?

marineam:
That sounds fine to me.

techdragon:
@marineam I'm a little stuck on the appropriate location in the work dir.

```src/third_party``` seems to only contain repos directly, so having a folder inside there that contains multiple git repositories seems like it could be confusing. however I'm not sure if creating a folder such as ```src/repo_configs``` makes much sense either.

Got any thoughts on this matter?

marineam:
It could also be located in the top level of the repo root instead of under src/ use a top level local/ or config/ directory. There are other configuration bits that currently scattered around the SDK such as /etc/shared_user_passwd.txt and src/scripts/.default_board. This new location could become the recommended place for that sort of stuff in the future.

techdragon:
@marineam a top level ```config``` dir sounds pretty hard to argue against. Thanks for the suggestion.  It shouldn't take too long to refactor if you want to check out how the logic looks.

techdragon:
New PR Created.

PR #371, opened 2015-01-05
Auto symlink repo configs

techdragon:
Better improved multiple repo support handling code. Replaces my prior pull requests coreos/scripts#370 and coreos/chromite#18

### DIFF
--- a/update_chroot
+++ b/update_chroot
@@ -87,6 +87,40 @@ location = ${COREOS_OVERLAY}
 location = ${PORTAGE_STABLE_OVERLAY}
 EOF
 
+# Automatically symlink any provided portage repo configurations into
+# the appropriate destination directory from "config/portage/repos".
+# CAUTON: Any repo configuration files with the same name, will be silently
+# relinked to the last file found.
+# Automatic file renaming to avoid this is challenging as it has to deal with
+# the case where we actually do want the files overwritten and so is outside
+# the scope of this function for now.
+# If your taking advantage of this functionality in the build toolchain
+# you probably found out about it from these code comments so we can assume
+# for now at least, that you have been somewhat warned.
+
+# If any directories match config/portage/repos/*/repos.conf
+if [[ -n $(sudo find /mnt/host/source/config/portage/repos \
+  -maxdepth 2 -mindepth 2 -type d -name "repos.conf") ]]; then
+  # If any files match config/portage/repos/*/repos.conf/*
+  if [[ -n $(sudo find /mnt/host/source/config/portage/repos \
+    -maxdepth 3 -mindepth 3 -path "*/repos.conf/*" -type f) ]]; then
+    info "Symlinking portage repo configuration files into repos.conf dir..."
+    # Do not name any of your repo configs "coreos.conf" bad things will happen.
+    # We abort if any file matches config/portage/repos/*/repos.conf/coreos.conf
+    if [[ -n $(sudo find /mnt/host/source/config/portage/repos \
+      -maxdepth 3 -mindepth 3 -path "*/repos.conf/*" -name "coreos.conf") ]]; then
+      die_notrace "One of the portage repo configuration files to be symlinked is " \
+        "named \"coreos.conf\" which will conflict with the main configuation file " \
+        "called \"coreos.conf\" which is created during the chroot build process."
+    fi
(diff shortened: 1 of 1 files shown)


update_chroot
```diff
@@ -87,6 +87,40 @@ location = ${COREOS_OVERLAY}
...
+# CAUTON: Any repo configuration files with the same name, will be silently
+# relinked to the last file found.
+# Automatic file renaming to avoid this is challenging as it has to deal with
+# the case where we actually do want the files overwritten and so is outside
+# the scope of this function for now.
+# If your taking advantage of this functionality in the build toolchain
+# you probably found out about it from these code comments so we can assume
+# for now at least, that you have been somewhat warned.
+
+# If any directories match config/portage/repos/*/repos.conf
```
marineam: Why bother with the multiple levels of directories? I think looking for `config/portage/repos.conf/*.conf` should be sufficient and has the advantage of mapping directly to `/etc/portage/repos.conf/`.

techdragon: @marineam I had that idea at first before I realised that prevents using multiple repo configs from different git repos.

I added the extra directory between ```config/portage``` and the target ```repos.conf/*.conf``` so that more than 1 git repo can be sourced using repo & its manifest format. It would be implemented without the extra directory level, but without that level you only get 1 source for those files. Neither the repo tools or vanilla git provide a way to pull in single files, so to avoid making the build script more rigid I added the extra directory level. This lets you add additional ```*.conf``` files into ```/etc/portage/repos.conf/``` just by adding a new project/repository to your manifest. For example, my ```repos.conf/cygnus.conf``` is stored in cygnus-labs/cygnus-repo-config and I can pull it into an otherwise completely vanilla current release build of CoreOS by creating a manifest file, and then specifying that manifest at the start of the build process. 
I select the manifest with this at the start of my Jenkins task.
````bash
export MANIFEST_VERSION=cygnus-test
````
Then later on in the Jenkins task just executes this to pull in everything to do with my extra portage overlay. 
````bash
repo init -u https://github.com/cygnus-labs/manifest.git -m ${MANIFEST_VERSION}.xml -g minilayout --repo-url https://chromium.googlesource.com/external/repo.git
````
After that everything is a pretty normal CoreOS build process other than the bits where I add extra packages.

The manifest ```cygnus-test.xml``` is very small and so far has let me significantly reduce how much I have "forked" in order to work with CoreOS.
````xml
<?xml version="1.0" encoding="UTF-8"?>
<manifest>
  <include name="release.xml" />

  <remove-project name="coreos/scripts" />
  <project path="src/scripts" name="cygnus-labs/scripts"
           revision="auto-symlink-repo-configs"
           groups="minilayout" />
  <project path="config/portage/repos/cygnus"
           name="cygnus-labs/cygnus-repo-config"
           groups="minilayout" />
  <project path="src/third_party/cygnus-overlay"
           name="cygnus-labs/cygnus-overlay"
           groups="minilayout" />
</manifest>
```` 

Thats the entire file, and i only have to have the remove-project for scripts and the project for scripts to add back in my own fork, so that I could test the branch with this Pull Request in it. In fact if this PR is merged, and if the repo tools don't have any trouble sourcing the include files if I make coreos/manifests a git subtree/submodule of the main repo. With that I'll be down to no 'forked' repos, just extra ones I'm pulling in as required and some extra steps in the build process which if skipped them turn my build into a completely normal CoreOS build. A little more documentation might help but other than that I'm not sure if the build process really could support adding extra ebuilds any easier way than this.
The extra directory level adds a lot of flexibility and at least to me seems better aligned with using the repo tool to manage all of the git repos involved in a CoreOS/ChromeOS project.

techdragon:
Is it possible to get some kind of commentary @marineam

I'd like to see it merged or critiqued, not sitting idle.

marineam:
@techdragon sorry, forgot about this. Could you squash the commits? I think the find-based scheme looks more complicated that it needs to be but I don't have a good recommendation at the moment so we can merge as-is and improve as-needed later.

techdragon:
ok, ill get this squashed, close this one and resubmit once i have a little time to clean it up.

crawford:
You don't have to close this PR. You can just force push the remote branch and it will update.

techdragon:
@marineam finally had time to squash this & push up a clean single commit PR

crawford:
Thanks. LGTM.

PR #373, opened 2015-01-09
catalyst: disable ccache and autoresume

marineam:
Disable ccache as it is causing issues in other builds so disable it
everywhere to be safe. Disable the autoresume feature because our build
process doesn't actually make use of it.


crawford:
LGTM

PR #374, opened 2015-01-16
build_library: add extract_docs()

crawford:
Related https://github.com/coreos/coreos-overlay/pull/1055

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -154,6 +154,15 @@ write_packages() {
     fi
 }
 
+extract_docs() {
+    local root_fs_dir="$1"
+
+    info "Extracting docs"
+    tar --create --auto-compress --file="${BUILD_DIR}/doc.tar.bz2" \
+        --directory="${root_fs_dir}/usr/share/coreos" doc
+    sudo rm --recursive --force "${root_fs_dir}/usr/share/coreos/doc"
+}
+
 # Add an entry to the image's package.provided
 package_provided() {
     local p profile="${BUILD_DIR}/configroot/etc/portage/profile"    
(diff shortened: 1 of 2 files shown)


build_library/build_image_util.sh
```diff
@@ -154,6 +154,15 @@ write_packages() {
     fi
 }
 
+extract_docs() {
+    local root_fs_dir="$1"
+
+    info "Extracting docs"
+    tar --create --bzip --file="${BUILD_DIR}/docs.tar" \
```
marineam: Missing a .bz2 extension. Also I think tar is the one case where I find long args more confusing than the short. This should do fine: `tar -xf "${BUILD_DIR}/doc.tar.bz2" -C "${root_fs_dir}/usr/share/coreos" doc`

That has the added bonus of including a top level directory in the tarball.

crawford: Doesn't this need to be `tar -caf "${BUILD_DIR}/doc.tar.bz2" -C "${root_fs_dir}/usr/share/coreos" doc`? The auto-compress flag isn't implied by default.

marineam: Ok, I think my broken example proved you were right to begin with. Leave it as-is :-P

marineam: except for the directory tweak and name change, best to always have a top level directory that matches the tarball name.

marineam: Oh, and probably remove extract_docs for the dev image, otherwise building a dev and prod image at the same time create conflicting tarballs.

crawford: Ha, ok. How about `tar --create --auto-compress --file="${BUILD_DIR}/docs.tar.bz2" --directory="${root_fs_dir}/usr/share/coreos" doc`?

crawford:
Updated.

build_library/build_image_util.sh
```diff
@@ -154,6 +154,15 @@ write_packages() {
     fi
 }
 
+extract_docs() {
+    local root_fs_dir="$1"
+
+    info "Extracting docs"
+    tar --create --auto-compress --file="${BUILD_DIR}/docs.tar.bz2" \
```
marineam: doc, not docs to match the dir :-P

marineam:
One lat nit and then LGTM

build_library/build_image_util.sh
crawford: doh

PR #376, opened 2015-02-06
oem/ami: fix ami_all.json

crawford:
This was not a valid json document before. Actually make this an object.


jedsmith:
belated lgtm!

PR #379, opened 2015-02-12
disk_util: allocate the maximum number of usable inodes

marineam:
This change changes the default 'bytes-per-inode' ration from 16K to 4K,
the block size. To prevent this from wasting too much space change the
inode size from the default 256 to the minimum size, 128. Larger inodes
are used to store extended attributes more efficiently but since we do
not use SELinux the majority of files do not have security attributes.

These defaults may be modified via the new `bytes_per_inode` and
`inode_size` options.

Fixes https://github.com/coreos/bugs/issues/264


crawford:
LGTM

PR #380, opened 2015-02-17
Updated VirtualBox script for new releases

skarllot:
Changes to handle stable, beta and alpha special versions.
Fixed to always import GPG key from official CoreOS web page.
Default version changed from alpha to stable.


marineam:
LGTM

PR #382, opened 2015-03-02
Enhanced dev_image_util with equivalent to --base_pkg

techdragon:
- "./build_image prod" already has the ability to specify which package will specify all the packages that should be pulled in and built into an image by specifying a package name using the --base_pkg command line flag. This creates an equivalent option for "./build_image dev" creating a --base_dev_pkg flag that passes a package name into the create_dev_img() function in dev_image_util.sh the same way that --base_pkg is passed into create_prod_image() inside prod_image_util.sh.


marineam:
LGTM, sorry I missed this earlier. Thanks!

PR #384, opened 2015-03-18
build_packages: add support breaking loops with multiple use flags

marineam:
Sorry, this is just getting bad. We need to switch to initializing the
board root stage3 style, similar to how the SDK or most any
semi-complicated Gentoo install starts. Breaking loops while merging
into a clean root is just too complicated.

On what this does: util-linux now has a udev *and* a systemd use flag.
Since we use systemd they both are effectively the same and pull in the
systemd package. This adds support for disabling both flags during the
loop breaking procedure.

### DIFF
--- a/build_packages
+++ b/build_packages
@@ -168,12 +168,23 @@ if [[ ${#CROS_WORKON_PKGS[@]} -gt 0 ]]; then
   )
 fi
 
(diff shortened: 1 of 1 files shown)


crawford:
LGTM

PR #386, opened 2015-03-25
disk_util: include fsck output during resize operation

marineam:
Previously fsck output was suppressed to reduce the amount of noise in
build logs on the assumption that fsck really shouldn't have a reason to
fail. The filesystem is freshly created after all. However some users
have reported that fsck is failing but without error messages we don't
know why.


brianredbeard:
Looks like a sensible solution. LGTM.

PR #387, opened 2015-03-26
oem/ami: update tests for etcd2

crawford:
The v1 API has been removed, so use v2 instead. The 10-second sleep was
added because the fleet tests were failing without it. My guess is that
etcd needed some time to warm up before we flooded it with requests.

### DIFF
--- a/oem/ami/test_ami.sh
+++ b/oem/ami/test_ami.sh
@@ -156,6 +156,9 @@ for host in ${ips[@]}; do
 done
 echo "OK ($instances)"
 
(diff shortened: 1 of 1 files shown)


marineam:
Hah! I knew we were going to forget about the v1 API in there.

Also ugg on the 10 second sleep... I'm sure whatever is going on there is going to show up in all sorts of places for lots of people.

LGTM

crawford:
Yeah, that sleep needs to be revisited...

PR #388, opened 2015-03-31
Fixes for cross builds

glevand:
Several minor fixes for when the target board is not amd64.  Please consider.

### DIFF
--- a/build_library/catalyst_toolchains.sh
+++ b/build_library/catalyst_toolchains.sh
@@ -16,10 +16,10 @@ configure_target_root() {
     local cross_chost=$(get_board_chost "$1")
     local profile=$(get_board_profile "${board}")
 
-    CHOST="${cross_chost}" \
-        ROOT="/build/${board}" \
-        SYSROOT="/usr/${cross_chost}" \
-        _configure_sysroot "${profile}"
+    CBUILD="$(portageq envvar CBUILD)" \
+        CHOST="${cross_chost}" \
+        ROOT="$ROOT" SYSROOT="$ROOT" \
+        _configure_sysroot "${CROSS_PROFILES[${cross_chost}]}"
 }
 
 build_target_toolchain() {
--- a/build_library/toolchain_util.sh
+++ b/build_library/toolchain_util.sh
@@ -203,8 +203,8 @@ _configure_sysroot() {
     $sudo tee "${ROOT}/etc/portage/make.conf" >/dev/null <<EOF
 $(portageq envvar -v CHOST CBUILD ROOT SYSROOT \
     PORTDIR PORTDIR_OVERLAY DISTDIR PKGDIR)
-HOSTCC=\${CBUILD}-gcc
-PKG_CONFIG_PATH="\${SYSROOT}/usr/lib/pkgconfig/"
+HOSTCC=${CBUILD}-gcc
+PKG_CONFIG_PATH="${SYSROOT}/usr/lib/pkgconfig/"
 EOF
 }
 
--- a/build_library/toolchain_util.sh
+++ b/build_library/toolchain_util.sh
@@ -187,7 +187,7 @@ _get_dependency_list() {
 
 # Configure a new ROOT
 # Values are copied from the environment or the current host configuration.
-# Usage: ROOT=/foo/bar SYSROOT=/foo/bar configure_portage coreos:some/profile
+# Usage: CBUILD=foo-bar-linux-gnu ROOT=/foo/bar SYSROOT=/foo/bar configure_portage coreos:some/profile
 _configure_sysroot() {
     local profile="$1"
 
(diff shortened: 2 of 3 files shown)


build_library/catalyst_toolchains.sh
```diff
@@ -16,10 +16,13 @@ configure_target_root() {
...
     local profile=$(get_board_profile "${board}")
 
+    # Get the host CBUILD.
+    local cbuild="$(portageq envvar CBUILD)"
+
     CHOST="${cross_chost}" \
         ROOT="/build/${board}" \
         SYSROOT="/usr/${cross_chost}" \
-        _configure_sysroot "${profile}"
+        _configure_sysroot ${cbuild} "${profile}"
```
marineam: It would be more in keeping with the existing (admittedly awkward) pattern to pass CBUILD via the environment. So something like:

```sh
CHOST="${cross_chost}" \
    CBUILD="$(portageq envvar CBUILD)" \
    ROOT="/build/${board}" \
    SYSROOT="/build/${board}" \
    _configure_sysroot "${profile}"
```

Any preferences? In this case _configure_sysroot wouldn't need to be modified.

glevand: Hi,

I pushed out a new version that uses an environment variable, and is also rebased to the latest.

With the environment variable, we need to set CBUILD before CHOST for $(portageq envvar CBUILD) to still pick up the host tuple.  Should we put a comment in about this?

Also, we need to modify _configure_sysroot regardless to write the CBUILD to the make.conf file.  See the new patch.

build_library/catalyst_toolchains.sh
```diff
@@ -16,10 +16,10 @@ configure_target_root() {
...
     local profile=$(get_board_profile "${board}")
 
-    CHOST="${cross_chost}" \
-        ROOT="/build/${board}" \
-        SYSROOT="/usr/${cross_chost}" \
-        _configure_sysroot "${profile}"
+    CBUILD="$(portageq envvar CBUILD)" \
+        CHOST="${cross_chost}" \
+        ROOT="$ROOT" SYSROOT="$ROOT" \
+        _configure_sysroot "${CROSS_PROFILES[${cross_chost}]}"
```
marineam: Why are you switching from the board profile  to to the generic CROSS_PROFILES? Does your board profile differ from the cross profile? Does your board profile not work for building a native toolchain?

Worthy of note on this stuff: In the case of amd64 we build the native toolchain with the intent to create developer images/containers that have a full native dev environment with emerge and everything. That goal may not be practical for other architectures, in which case providing glibc and gcc's libstdc++.so/libgcc_s.so by other means may be reasonable. ChromiumOS handled this by just copying the cross compiler's libs instead of building a full toolchain like I do now.

build_library/catalyst_toolchains.sh
marineam: Feel free to comment on the ordering considering I got it wrong, I had misunderstood the order in which the shell sets the variables and evaluates `"$(portageq envvar CBUILD)"` in my example.

build_library/toolchain_util.sh
```diff
@@ -201,10 +201,10 @@ _configure_sysroot() {
     $sudo eselect profile set --force "$profile"
 
     $sudo tee "${ROOT}/etc/portage/make.conf" >/dev/null <<EOF
-$(portageq envvar -v CHOST CBUILD ROOT SYSROOT \
-    PORTDIR PORTDIR_OVERLAY DISTDIR PKGDIR)
-HOSTCC=\${CBUILD}-gcc
-PKG_CONFIG_PATH="\${SYSROOT}/usr/lib/pkgconfig/"
+$(portageq envvar -v CHOST ROOT SYSROOT PORTDIR PORTDIR_OVERLAY DISTDIR PKGDIR)
```
marineam: Since portageq will inherit CBUILD from the environment this should work as it was, but the \$ escapes need to remain to make that work so the vars in HOSTCC and such are interpreted later by portage.

Is that to magical?

build_library/catalyst_toolchains.sh
marineam: Another odd thing in this diff, ROOT needs to point to the target boards root, SYSROOT needs to point to the cross-compiler's sysroot to provide the initial dependencies of the toolchain, they don't yet exist in ROOT. I presume this change was coppied from elsewhere where ROOT and SYSROOT were indeed the same.

marineam: Ohhh, I see the whole `"${CROSS_PROFILES[${cross_chost}]}"` thing got copied from there too. Now this all makes sense.

marineam:
I trimmed down the diff a little and reposted this as #390, let me know if I missed something.

marineam:
This also may help: https://github.com/coreos/scripts/pull/391

PR #389, opened 2015-04-01
catalyst: hacky fix for amd64 toolchain builds

marineam:
The mpc upgrade is breaking the board toolchain build because the
two-stage build process is improperly kicking in, the wrong libraries
get linked, sysroot ignored, and other unpleasentries.

This should work around the issue pending fixing the gcc ebuild or
whatever is allowing the two-stage build to kick in for cross-compiled
builds.


marineam:
Nope, still wrong... investigating more

marineam:
Oh, I'm a dummy, this is the CBUILD bug.

PR #390, opened 2015-04-01
Fixes for cross compiling

marineam:
Replaces https://github.com/coreos/scripts/pull/388, pruning down the diff a little.

In addition to (hopefully) fixing true cross-architecture builds this resolves an issue caused by the recent mpc upgrade. Since without this CHOST==CBUILD cross-compiling logic wasn't properly engaged and the two-phase gcc build kicked in, linking against some things under `/build/$BOARD` but then failing to run because the host chroot had a different mpc library.


crawford:
LGTM

glevand:
I rebased to the latest (10c9a82) and things seems to work OK for my arm64 port.

PR #392, opened 2015-04-07
Add basic support for arm64 

glevand:
Adds basic support for arm64 platforms.  Please consider.

-Geoff

### DIFF
--- a/build_library/toolchain_util.sh
+++ b/build_library/toolchain_util.sh
@@ -15,6 +15,7 @@ TOOLCHAIN_PKGS=(
 # compiler to build a full native toolchain. Packages are not uploaded.
 declare -A CROSS_PROFILES
 CROSS_PROFILES["x86_64-cros-linux-gnu"]="coreos:coreos/amd64/generic"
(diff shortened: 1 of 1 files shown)


build_library/toolchain_util.sh
```diff
@@ -38,7 +43,7 @@ declare -r \
 # Usage: get_portage_arch chost
 get_portage_arch() {
     case "$1" in
-        aarch64*)   echo arm;;
+        aarch64*|arm64*) echo arm64;;
```
marineam: Is arm64 a valid CHOST prefix? The list that this was copied from (which has since been updated for arm64 in Gentoo) doesn't include it:
```sh
case ${CTARGET} in
                aarch64*)   TARCH=arm64;;
                alpha*)     TARCH=alpha;;
                arm*)       TARCH=arm;;
                hppa*)      TARCH=hppa;;
                ia64*)      TARCH=ia64;;
                i?86*)      TARCH=x86;;
...
```

glevand:
Hi, No, arm64 is not a valid CHOST.  I set this up by just reading the code, and missed how get_board_chost() works.  I pushed out an updated version.

-Geoff

marineam:
LGTM, thanks!

PR #396, opened 2015-04-13
Add support for Secure Boot images and sign unofficial builds with test ...

mjg59:
...keys

Add qemu_uefi_secure target for building Secure Boot images. These are
identical to qemu_uefi images with the exception that the test keys have
been installed into the flash image, enabling Secure Boot by default. In
addition, sign the grub binary with the test keys during build when producing
unofficial images.

### DIFF
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -141,8 +141,17 @@ case "${FLAGS_target}" in
     x86_64-efi)
         info "Installing default x86_64 UEFI bootloader."
         sudo mkdir -p "${ESP_DIR}/EFI/boot"
-        sudo cp "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}" \
-            "${ESP_DIR}/EFI/boot/bootx64.efi"
+	# Use the test keys for signing unofficial builds
+	if [[ ${COREOS_OFFICIAL:-0} -ne 1 ]]; then
+            sudo sbsign --key /usr/share/sb_keys/DB.key \
+		--cert /usr/share/sb_keys/DB.crt \
+                    "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}"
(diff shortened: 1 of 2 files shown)


build_library/grub_install.sh
```diff
@@ -143,6 +143,13 @@ case "${FLAGS_target}" in
         sudo mkdir -p "${ESP_DIR}/EFI/boot"
         sudo cp "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}" \
             "${ESP_DIR}/EFI/boot/bootx64.efi"
+	# Use the test keys for signing unofficial builds
+	if [[ ${COREOS_OFFICIAL:-0} -ne 1 ]]; then
+            sudo sbsign --key /usr/share/sb_keys/DB.key \
```
marineam: Perhaps sign `"${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}"` instead (it expands to `"${ESP_DIR}/coreos/grub/x86_64-efi/core.efi"`) Then in that directory there is both a core.efi and a core.efi.signed which may be a useful reference to someone digging around trying to understand the boot process. It is less obvious when it is core.efi that is unsigned but bootx64.efi which lives elsewhere that is signed.

build_library/vm_image_util.sh
```diff
@@ -590,6 +595,20 @@ _write_qemu_uefi_conf() {
     VM_GENERATED_FILES+=( "$(_dst_dir)/${ovmf_ro}" "$(_dst_dir)/${ovmf_rw}" )
 }
 
+_write_qemu_uefi_secure_conf() {
+    local tmpdir=`mktemp -d`
```
marineam: You can just use `${VM_TMP_DIR}` instead of creating a new dir, that way failures don't leave things scattered around.

mjg59:
Reworked to match feedback.

marineam:
LGTM!

PR #397, opened 2015-04-16
Add support for signed kernels

mjg59:
grub needs shim in order to verify kernels, and we need to rework our image format a little to enable the dm-verity work. This gets us to a fully trusted boot chain in the qemu-uefi-secure target.

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -1,5 +1,8 @@
 # Main GRUB config
 
+# Set the prefix back to the correct value after we're done with memdisk
+set prefix=($root)/coreos/grub
+
 # Load any and all video drivers.
 # Required under UEFI to boot Linux with a working console.
 insmod all_video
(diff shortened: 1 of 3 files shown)


build_library/grub.cfg
```diff
@@ -47,28 +50,34 @@ function load_coreos {
         set suf=""
     fi
 
-    echo "Loading ($root)/boot/vmlinuz"
-    linux$suf /boot/vmlinuz $linux_console $linux_root "$@" $linux_append
+    search --no-floppy --set root --part-label EFI-SYSTEM
```
marineam: $root is initially set th the ESP in load.cfg:

```
search.fs_uuid ${ESP_FSID} root \$root
```

the gptprio.next calls should be updated to set $usr or similar instead of $root

build_library/grub.cfg
```diff
@@ -47,28 +50,34 @@ function load_coreos {
...
-    fi
+    echo "Loading ($root)$kernel"
+    linux$suf $kernel $linux_console $linux_root "$1" $linux_append
 
     echo "Booting CoreOS!"
 }
 
 menuentry "CoreOS default" --id=coreos {
     gptprio.next -d root -u usr_uuid
+    if [ "$root" = "hd0,gpt3" ]; then
```
marineam: Assuming the name 'hd0' isn't going to be reliable, also I'd like to avoid making assumptions about our partition numbers even if the assumption is safe today. Perhaps we name kernel based on partition UUID or label so there isn't a translation from gptprio.next or if you do translate the UUIDs are currently static:

USR-A: 7130c94a-213a-4e5a-8e26-6cce9662f132
USR-B: e03dd35c-7c2d-4a47-b3fe-27f15780a57c

These are defined in scripts/build_library/disk_layout.json

Using the label USR-A or USR-B would look prettier but that requires adding a new option to the gptprio.next command in grub/grub-core/commands/gptprio.c

build_library/grub.cfg
```diff
@@ -38,7 +41,7 @@ if [ -z "$linux_console" ]; then
 fi
 
 
-# Load a kernel and boot! $root must point at USR-A or USR-B
+# Load a kernel and boot!
 function load_coreos {
     # UEFI uses linuxefi/initrdefi instead of linux/initrd
```
marineam: if we move the logic to select between the linux and linuxefi commands to the top level the load_coreos function could just go away.

mjg59:
Updated based on comments

marineam:
Looks good, before this lands we need to revive some of the kernel copying goo in here: https://github.com/coreos/update_engine/blob/master/coreos-postinst

basically move the ESP mounting code outside the if, and use the old cros_legacy flag to determine if the old syslinux config/kernels need to be copied or if this new kernel location should be used. I'm fine with copying the kernels over to the ESP even if it is a grub version that will ignore them to reduce the conditional logic required.

marineam:
Or, we just wait to merge this until after update_engine has support for our new kernel payloads. That reduces churn in coreos-postinst but means this PR will linger for a while longer.

mjg59:
https://github.com/coreos/update_engine/pull/67 is intended to cover this for now. Obviously this won't work in the dm-verity world so it's just a temporary solution.

marineam:
[LGTM](http://www.lgtm.in/i/qtqgy0jOp)

PR #399, opened 2015-04-17
setup_board: explicitly perform all package moves

marineam:
For some reason package moves are not handled automatically in the board
build roots. Add explicit calls to emaint to update cached binary
packages, installed packages, and the world file.


crawford:
LGTM

PR #400, opened 2015-04-20
oem/ami: add new import script, replaces build_ebs_on_ec2.sh

marineam:
This script uses the EC2 volume import tools instead of attaching and
writing to an EBS volume. This mechanism will be useful for creating
AMIs in isolated EC2 regions and can be run from any host with API
access and the EC2 tools.

TODO: Allow region to be specified and automatically create region-local
S3 buckets as needed. This version hard codes a bucket only usable by
our dev AWS account, not prod. Later on: move to a more compact disk
format like VMDK.

### DIFF
--- a/oem/ami/import.sh
+++ b/oem/ami/import.sh
@@ -1,0 +1,203 @@
+#!/bin/bash
+#
+# This expects to run on an EC2 instance.
+#
+# mad props to Eric Hammond for the initial script
+#  https://github.com/alestic/alestic-hardy-ebs/blob/master/bin/alestic-hardy-ebs-build-ami
+
+# Set pipefail along with -e in hopes that we catch more errors
+set -e -o pipefail
+
+DIR=$(dirname $0)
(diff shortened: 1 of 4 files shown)


oem/ami/import.sh
```diff
@@ -0,0 +1,204 @@
...
+    TODAYS_VERSION=$(( (`date +%s` - ${COREOS_EPOCH}) / 86400 ))
+    VERSION="${TODAYS_VERSION}-$(date +%H-%M)"
+    GROUP="master"
+fi
+
+# Size of AMI file system
+# TODO: Perhaps define size and arch in a metadata file image_to_vm creates?
+size=8 # GB
+arch=x86_64
+arch2=amd64
```
crawford: How is this used?

marineam: It isn't, just happened to be in the old build script I copied from.

crawford:
LGTM

PR #401, opened 2015-04-21
oem/ami: make import zone and bucket conferable

marineam:
The intent is to export the bucket and optionally the zone in the
environment alone with the AWS auth credentials.


crawford:
LGTM

PR #402, opened 2015-04-21
vm_image_util: add support for VMDK images for EC2 AMIs

marineam:
Lets us avoid larger data transfers when importing images.


crawford:
LGTM

PR #403, opened 2015-04-23
oem/ami: update path to ami scripts directory

marineam:
The ami builder host now has a proper checkout of the scripts repo.


crawford:
LGTM

PR #404, opened 2015-04-24
oem/ami: add support for publishing AMIs in us-gov-west-1

marineam:
A few general changes:
 - Use https for EC2 endpoint URLs
 - Remove parallelism from prod-publish, was launching enough Java
   processes at once to trigger OOM :(
 - Only share snapshots with Amazon in us-east-1: only needed for
   marketplace listing and marketplace only uses us-east-1.


crawford:
LGTM

PR #405, opened 2015-04-30
ROADMAP: add initial overall CoreOS roadmap document

marineam:
This doc covers CoreOS as a whole and a few details that don't really
fit anywhere else. Individual projects maintain their own roadmaps.

Update LICENSE and add a README as general housekeeping.


crawford:
Sounds good.

PR #407, opened 2015-05-04
scripts: Extend import.sh to create AMIs with an encrypted root device

### LINKED ISSUE #406: Allow for encrypted AMI volumes
There is no easy way to have an EC2 instance with an encrypted volume when booting up from the CoreOS provided AMIs.  I suggest that we modify the [build script](https://github.com/coreos/scripts/blob/master/oem/ami/build_ebs_on_ec2.sh) for AMIs so that an additional AMI is built per virtualization type with an encrypted volume.  

There is a reduced list of instances available that can run the additional AMIs with encryption enabled, as not all instances support EBS encryption.

KnownSubset:
An additional AMI is registered per virtualization type.
This is necessary as AWS offers no way to go from an unencrypted
snapshot to encrypted or vice-versa.

Fixes #406

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
 Copyright (c) 2006-2013 The Chromium OS Authors. All rights reserved.
-Copyright (c) 2013 The CoreOS Authors. All rights reserved.
+Copyright (c) 2013-2015 CoreOS, Inc. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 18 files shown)


marineam:
The purpose of `import.sh` was to drop the need to attach EBS volumes. You are welcome to contribute a script for creating encrypted AMIs but I would recommend doing so as an option to the older `build_ebs_on_ec2.sh` or as an entirely new script based on it.

PR #408, opened 2015-05-11
Don't read OEM config data if Secure Boot is enabled

mjg59:
We don't want untrusted configuration to be read if we're in Secure Boot
mode, so skip the OEM config when Secure Boot is enabled and in User Mode.

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -18,11 +18,22 @@ set linux_console=""
 # Anything else the OEM adds should use this variable.
 set linux_append=""
 
+set secure_boot="0"
+
+if [ "$grub_platform" = "efi" ]; then
+   getenv -e SecureBoot -g 8be4df61-93ca-11d2-aa0d-00e098032b8c -b sb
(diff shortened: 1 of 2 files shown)


build_library/grub.cfg
```diff
@@ -18,11 +18,22 @@ set linux_console=""
...
+   if [ "$sb" = "01" -a "$setupmode" = "00" ]; then
+      set secure_boot="1"
+   fi
+fi
 
 # Search for the OEM partition, load additional configuration if found.
-search --no-floppy --set oem --part-label OEM --hint "$root"
-if [ -n "$oem" -a -f "($oem)/grub.cfg" ]; then
-    source "($oem)/grub.cfg"
+if [ "$sb" = "0" ]; then
```
marineam: I presume this should be `$secure_boot`

mjg59: Oops, yes.

marineam:
[LGTM](http://lgtm.in/i/qT1t0zHQp)

PR #419, opened 2015-06-10
core_promote: update to new standard GCE image description format

marineam:
Description will now look like:

    CoreOS, CoreOS alpha, 695.0.0, amd64-usr published on 2015-05-28


crawford:
LGTM

glevand:
Is this image description format documented somewhere?

marineam:
@glevand not that I'm aware. It is only applicable to images included in the global "standard" list everyone sees by default, Google requested it so they could improve how the list is displayed.

PR #420, opened 2015-06-18
recognize ssh ed25519 keys

ibukanov:
This extends in scripts/contrib/create-basic-configdrive the regexp matching ssh rsa and dss keys to ed25519 keys so modern crypto is supported.


marineam:
LGTM

PR #421, opened 2015-06-18
dm-verity build support

gtank:
This adds dm-verity support to prod builds under the flag `--enable_verity`. It currently has an issue with preventing the OEM disk from mounting.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -258,11 +258,11 @@ finish_image() {
 
   local disk_img="${BUILD_DIR}/${image_name}"
 
+  # Copy kernel to support dm-verity boots
   sudo mkdir -p "${root_fs_dir}/boot/coreos"
   sudo cp "${root_fs_dir}/usr/boot/vmlinuz" \
        "${root_fs_dir}/boot/coreos/vmlinuz-a"
-  sudo cp "${root_fs_dir}/usr/boot/vmlinuz" \
-       "${root_fs_dir}/boot/coreos/vmlinuz-b"
+
   # Record directories installed to the state partition.
   # Explicitly ignore entries covered by existing configs.
   local tmp_ignore=$(awk '/^[dDfFL]/ {print "--ignore=" $2}' \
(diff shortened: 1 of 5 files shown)


build_library/grub_install.sh
```diff
@@ -118,6 +123,16 @@ EOF
 
 if [[ ! -f "${ESP_DIR}/coreos/grub/grub.cfg.tar" ]]; then
     info "Generating grub.cfg memdisk"
+
+    # Hack around our need for incompatible default behaviors in different image types
+    if [[ ${FLAGS_prod} -eq ${FLAGS_TRUE} ]]; then
+      # use dm-verity for /usr
+      cp ${BUILD_LIBRARY_DIR}/grub-prod.cfg ${BUILD_LIBRARY_DIR}/grub.cfg
```
vcaputo: quotes

build_library/grub_install.sh
```diff
@@ -71,6 +73,9 @@ cleanup() {
     if [[ -b "${LOOP_DEV}" ]]; then
         sudo losetup --detach "${LOOP_DEV}"
     fi
+    if [[ -e "${BUILD_LIBRARY_DIR}/grub.cfg" ]]; then
```
marineam: wat?!?

build_library/disk_util
```diff
@@ -582,7 +582,7 @@ def Update(options):
 
 
 def Mount(options):
-  """Mount the given disk image.
+  """Mount the whole given disk image or (optionally) a single labeled partition from it.
```
marineam: How about we just create a new command that mounts only the given partition to a specific path, skipping the extra mount directory and bind stuff that only really makes sense for full disk images. To do that split the little bit of common code into a shared function

```python
     mount_opts = ['loop',
                   'offset=%d' % mount['image_first_byte'],
                   'sizelimit=%d' % mount['image_bytes']]
     if options.read_only:
       mount_opts.append('ro')
     elif (mount.get('fs_type', None) in ('ext2', 'ext4') and
           not IsE2fsReadWrite(options, mount)):
       mount_opts.append('ro')
 
     if mount.get('fs_subvolume', None):
       mount_opts.append('subvol=%s' % mount['fs_subvolume'])
 
     Sudo(['mkdir', '-p', full_path])
     Sudo(['mount', '-t', mount.get('fs_type', 'auto'),
                    '-o', ','.join(mount_opts),
                    options.disk_image, full_path])
```

build_library/disk_util
```diff
@@ -637,13 +642,25 @@ def Mount(options):
   for mount in sorted(mounts, key=len):
     DoMount(mounts[mount])
 
+
 def Umount(options):
-  """Unmount the given path.
```
marineam: Once the mount command is split to just mount to a specific directory instead of interpreting `part['mount']` this function can be left as-is

build_library/grub-prod.cfg
```diff
@@ -0,0 +1,84 @@
...
+    set suf="efi"
+else
+    set suf=""
+fi
+
+menuentry "CoreOS default" --id=coreos {
+    gptprio.next -d usr -u usr_uuid
+    if [ "$usr_uuid" = "7130c94a-213a-4e5a-8e26-6cce9662f132" ]; then
+       linux$suf /coreos/vmlinuz-a $linux_console $linux_root \
+            mount.usr=/dev/mapper/usr \
```
marineam: What about something like `@@MOUNTUSR@@=PARTUUID=$usr_uuid` and then just use sed to replace `@@MOUNTUSR@@` with `mount.usr` or `mount.usr=/dev/mapper/usr verity.usr`

Then we don't wind up with two nearly identical files to maintain.

build_library/grub_install.sh
gtank: it's cleanup related to the nasty dual-grub.cfg hack

build_library/prod_image_util.sh
```diff
@@ -76,7 +77,7 @@ create_prod_image() {
 L+  /etc/ld.so.conf     -   -   -   -   ../usr/lib/ld.so.conf
 EOF
 
-  # Only try to disable rw on /usr if there is a /usr partition 
+  # Only try to disable rw on /usr if there is a /usr partition
```
marineam: How about we add some 'read-only' flag to finish_image so there is no need to mount the ESP again. That would avoid all those changes to disk_util too.

build_library/grub_install.sh
marineam: I don't follow. This file is in git

build_library/prod_image_util.sh
gtank: Running `grub_install.sh` at the end of finish_image requires the ESP to be unmounted first. Does it make sense to move that to another part of the process, or change how it works? Avoiding that unmount does remove the need for any disk_util changes.

build_library/grub_install.sh
gtank: This commit renames the default grub.cfg to grub-dev.cfg, and copies the correct variant for the image type to grub.cfg so that `tar` still works for memdisk creation. I strongly prefer your suggestion above.

marineam: oh, I see.... even if you stick with two files you need to use a temp directory, you should absolutely not be modifying files in git.

build_library/prod_image_util.sh
marineam: I don't follow the concern, you would basically be copying kernel and signing just as finish_image already does, only difference is adding the verity and dd commands to finish_image and removing from here.

build_library/build_image_util.sh
```diff
@@ -294,33 +294,57 @@ finish_image() {
   # Zero all fs free space to make it more compressible so auto-update
   # payloads become smaller, not fatal since it won't work on linux < 3.2
   sudo fstrim "${root_fs_dir}" || true
-  if mountpoint -q "${root_fs_dir}/usr"; then
```
marineam: need to keep this if

marineam: and since a lot of the following shouldn't happen without a /usr partition it may need to move to a new function to keep things clean, nesting things under a big if could work too, whatever looks better.

the only time there isn't a /usr partition is when running `./build_image container` which creates a dev image with a single partition, useful for passing to nspawn (not useful for booting)

build_library/build_image_util.sh
```diff
@@ -294,33 +294,57 @@ finish_image() {
...
-      sudo mv "${root_fs_dir}/boot/coreos/vmlinuz-b.signed" \
-	   "${root_fs_dir}/boot/coreos/vmlinuz-b"
   fi
+
   rm -rf "${BUILD_DIR}"/configroot
   cleanup_mounts "${root_fs_dir}"
   trap - EXIT
 
+  # Check whether we want to use verity boot options
+  local enable_verity=${FLAGS_enable_verity}
```
marineam: just use add FLAGS_ to the vars below. it isn't *that* much longer.

build_library/grub_install.sh
```diff
@@ -118,8 +123,22 @@ EOF
 
 if [[ ! -f "${ESP_DIR}/coreos/grub/grub.cfg.tar" ]]; then
     info "Generating grub.cfg memdisk"
+
+    grub_temp_dir=$(mktemp -d)
```
marineam: nothing cleans this up

marineam: easiest way would be to make it a global var above and add to the cleanup function

marineam:
LGTM but looks like you need to rebase

marineam:
LGTM :)

PR #422, opened 2015-06-23
Add check for mkisofs

mrekucci:
This check helps to avoid exit the script with the message indicating
success even though the mkisofs is missing and an image has not been created.

### DIFF
--- a/contrib/create-basic-configdrive
+++ b/contrib/create-basic-configdrive
@@ -55,6 +55,12 @@ if [ $(id -u) -eq 0 ]; then
     exit 1
 fi
 
+# mkisofs tool check
+if ! which mkisofs &>/dev/null; then
+    echo "$0: mkisofs tool is required to create image." >&2
+    exit 1
+fi
+
 if [ -z "$HNAME" ]; then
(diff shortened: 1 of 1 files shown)


contrib/create-basic-configdrive
```diff
@@ -55,6 +55,13 @@ if [ $(id -u) -eq 0 ]; then
     exit 1
 fi
 
+# mkisofs tool check
+which mkisofs &>/dev/null
+if [ $? -ne 0 ]; then
```
crawford:
```sh
if which mkisofs &>/dev/null; then
```

mrekucci: Ok, I'll change it. I tried to be consistent, you have this style in `create-coreos-vdi` script when checking for VirtualBox tools.

mrekucci: It doesn't work that way you wrote it.

marineam: @crawford's example was just inverted:

```sh
if ! which mkisofs &>/dev/null; then
```

marineam:
LGTM, thanks!

PR #423, opened 2015-06-25
Adding packet to the list of img and oem types.

stresler:
Adding packet to the list of img and oem types.

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -36,6 +36,7 @@ VALID_IMG_TYPES=(
     secure_demo
     niftycloud
     cloudsigma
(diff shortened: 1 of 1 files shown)


crawford:
You'll at least need to define `IMG_packet_OEM_PACKAGE=oem-packet`. This will correspond to an ebuild in https://github.com/coreos/coreos-overlay/tree/master/coreos-base.

stresler:
I've added that.  I'll look in to the other available constants as well, but I think this is good for now. Thanks.

marineam:
LGTM, will merge after https://github.com/coreos/coreos-overlay/pull/1282

PR #424, opened 2015-06-25
build_packages: always rebuild when breaking dependency cycles

marineam:
Previously we attempted to speed up the first build of a board by
pulling in binary packages regardless of their use flags, assuming that
if the package was already built the loop would not be an issue. Usually
this was true but not always. Now things like util-linux will always get
compiled when setting up a board for the first time but at least it
won't be as likely to fail.


glevand:
I've noticed this problem before, so seems like a good thing.

mischief:
lgtm

PR #425, opened 2015-06-27
Build selinux policy

mjg59:
Create an selinux policy at image build time - despite the -i flag, this
won't attempt to install it into the running kernel.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -298,6 +298,11 @@ finish_image() {
     sudo fstrim "${root_fs_dir}/usr" || true
   fi
 
+  # Build the selinux policy
+  if [[ "$BOARD" = amd64* ]]; then
+      sudo chroot ${root_fs_dir} bash -c "cd /usr/share/selinux/mcs; semodule -i *.pp"
+  fi
+
   # Sign the kernels after /usr is in a consistent state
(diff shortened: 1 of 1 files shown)


build_library/build_image_util.sh
```diff
@@ -298,6 +298,9 @@ finish_image() {
     sudo fstrim "${root_fs_dir}/usr" || true
   fi
 
+  # Build the selinux policy
+  sudo chroot ${root_fs_dir} bash -c "cd /usr/share/selinux/mcs; semodule -i *.pp"
```
marineam: If this isn't practical to do in a cross friendly manner right now it'll need to be wrapped in something like `if [[ "$BOARD" = amd64* ]]` to keep arm happy.

mjg59: Oh hm. Yeah that's going to be a pain. I'll wrap it for now.

marineam: it's not the only thing that is going to suck for moving to other archs, dracut also screws us :(

mjg59: "Standard" approach is qemu and binfmt_misc…

marineam: Joy ;-)

marineam:
LGTM

PR #426, opened 2015-06-29
bootstrap_sdk: add support for directly GPG signing SDK tarballs

marineam:
SDK tarballs have a .DIGESTS file but it is created by catalyst instead
of the upload_image function. In order to support plain GPG signing but
not avoid re-generating .DIGESTS we need to move that code out of
upload_image to a new function. upload_files shouldn't do it itself
because it is also used for portage binary packages which shouldn't be
signed (there is no point, nothing would verify the signatures).


crawford:
LGTM

PR #427, opened 2015-07-01
disk_util: use the dummy disk UUID

crawford:
Depends on https://github.com/coreos/seismograph/pull/32.


vcaputo:
lgtm

marineam:
should this wait until there is a job in the initrd to initialize it on first boot?

crawford:
@marineam yes, I should have called that out.

crawford:
Depends on https://github.com/coreos/bootengine/pull/49.

PR #433, opened 2015-07-08
build_packages: be less aggressive about rebuilding packages

marineam:
The --rebuild-if-unbuilt emerge option will recompile packages if any
build-time dependencies changed. This can be a useful safety to make
sure applications get re-linked and so forth. However recent versions of
portage have grown better defaults for choosing when to rebuild so lets
turn the build_packages flag around and leave emerge to do its default
behavior and see if that is sufficiently safe.


crawford:
LGTM

PR #436, opened 2015-07-16
Revert "build_library: set osType to otherlinux-64"

marineam:
This reverts commit 39bb800f16afd077c3e7d930450bf327a6d895db.

This change disabled a number of features so it isn't suitable for the
generic VMware templates. We need to re-trace our steps to list exactly
what tools/systems weren't accepting the linux26 type.


crawford:
LGTM

oliverkurth:
Great, thank you :-)

PR #437, opened 2015-07-17
grub_install: stash an exact copy of the MBR boot code.

marineam:
Useful for repairing the MBR without needing GRUB's tools.


crawford:
Nice

PR #438, opened 2015-07-24
update_chroot/build_packages: remove obsolete packages

marineam:
Before attempting to do @preserved-rebuild to fix linked against old
libraries remove any packages that no longer have corresponding ebuilds,
making them impossible to rebuild. This uses `--depclean`'s secondary
meaning: `--unmerge` but only remove packages without dependencies.


crawford:
LGTM

vcaputo:
lgtm, if it works :)

PR #439, opened 2015-07-30
check_root: update ignore list for arm64, make error non-fatal

marineam:


glevand:
build_packages for arm64 now works (with warnings) with this patch applied.

LGTM

PR #442, opened 2015-08-05
scripts: Run ldconfig in qemu

glevand:
ldconfig does not work for non-native arches.  Create a new
build_image routine run_ldconfig that uses qemu user emulation
to run the board ldconfig on the board rootfs when the board and
SDK arches are different.

See: http://code.google.com/p/chromium/issues/detail?id=378377

Prior to calling run_ldconfig the board rootfs must have ldconfig
installed.  To arrange this move the call of run_ldconfig to after
the base package install.

Fixes build_image errors like these when building for arm64:

  /sbin/ldconfig: /lib64/libXXX is for unknown machine 183.

Depends on https://github.com/coreos/coreos-overlay/pull/1424

@marineam I'm still testing this, but wanted to get it out for your comments before I put too much into it.
/cc @andrejro

### DIFF
--- a/build_library/dev_image_util.sh
+++ b/build_library/dev_image_util.sh
@@ -96,6 +96,7 @@ create_dev_image() {
 
   set_image_profile dev
   emerge_to_image "${root_fs_dir}" @system ${base_pkg}
+  run_ldconfig "${root_fs_dir}"
   write_packages "${root_fs_dir}" "${BUILD_DIR}/${image_packages}"
   write_licenses "${root_fs_dir}" "${BUILD_DIR}/${image_licenses}"
 
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -93,6 +93,21 @@ generate_update() {
   upload_image -d "${update}.DIGESTS" "${update}".{bin,gz,zip}
 }
 
+# ldconfig cannot generate caches for non-native arches.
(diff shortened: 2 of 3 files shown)


build_library/dev_image_util.sh
```diff
@@ -96,6 +96,7 @@ create_dev_image() {
 
   set_image_profile dev
   emerge_to_image "${root_fs_dir}" @system ${base_pkg}
+  run_ldconfig "${root_fs_dir}"
```
marineam: Why not put this in emerge_to_image after the env-update call?

glevand: @marineam We call emerge_to_image twice.  First in start_image for sys-apps/baselayout, and later for the ${base_pkg}. For the start_image call ldconfig is not installed yet, so run_ldconfig will fail for arm64 there.

marineam: ah, ok.

marineam:
LGTM, it seems like a cross-toolchain bug that there isn't a $CHOST-ldconfig available but chances are that's a can of worms.

glevand:
@marineam when looking into the fix it seems that there once was such support, but was removed.  

Also, you'll need to merge this as I don't have commit rights to the scripts repo.  I could build amd64-usr dev and prod images, so I think this is OK to merge.

marineam:
done :)

PR #447, opened 2015-08-24
Ensure that selinux policy is installed in correct location

mjg59:
Failing to explicitly set the selinux policy store to operate on may
result in semodule installing the policy in an incorrect location. Pass
it on the command line in order to avoid this.


marineam:
LGTM

PR #448, opened 2015-09-01
grub: run Ignition on first boot

crawford:


marineam:
LGTM

marineam:
A slightly different theory of operation that may be useful: instead of an ignition specific kernel args we use `coreos.first_boot=1` and we use that to trigger ignition and force the removal of /sysroot/etc/machine-id if it exists to ensure systemd's first-boot logic also triggers. ConditionFirstBoot= may become more useful to us then. (I doubt ConditionFirstBoot works in the initrd though, it would just be there for full-system-boot-time things)

vcaputo:
This is cool.

PR #450, opened 2015-09-01
grub: add coreos.first_boot kernel param

crawford:


build_library/grub.cfg
```diff
@@ -37,11 +37,11 @@ if [ "$secure_boot" = "0" ]; then
    fi
 fi
 
-# Determine if this is a first boot. Run Ignition if so.
+# Determine if this is a first boot.
 search --no-floppy --set first_boot \
        --disk-uuid 00000000-0000-0000-0000-000000000001
 if [ -n "$first_boot" ]; then
-    set ignition="coreos.ignition.run=1"
+    set coreos="coreos.first_boot=1"
```
crawford: Would it be useful to pass along the first-boot disk?

`set coreos="coreos.first_boot=$first_boot"`

marineam: perhaps except that the device name in GRUB will have nothing to do with the device name in the kernel so not that useful right now.

I think it would be worth while some day to make sure the boot device (with the ESP partition) is recorded and more strictly required to be the device that we look at while booting, similar to the behavior defined by the systemd discoverable partition spec. That way you don't risk conflicts if you plug two coreos disks into the same box (to recover data for example) because only the disk the firmware boots will actually be booted. however there are a lot of details to make that work and we are a long ways away so no point in worrying about it right now.

crawford: OK, I'll leave it as is then.

build_library/grub.cfg
```diff
@@ -11,10 +11,10 @@ insmod all_video
 set default="coreos"
 set timeout=1
 
-# Default kernel args for root filesystem, console, and Ignition.
+# Default kernel args for root filesystem, console, and CoreOS.
 set linux_root="root=LABEL=ROOT"
 set linux_console=""
-set ignition=""
+set coreos=""
```
marineam: maybe first_boot would be a better name

crawford: Sure.

marineam:
LGTM

PR #452, opened 2015-09-03
add windows tools for running vms on virtualbox

fivethreeo:
this adds powershell scripts with automatic tool downloads for making a virtualbox vdi and a configdrive for running coreos

missing: gpg  checks of coreos image, destinaton dir for output files, 
issues: download problems from s3 with bitstransfer, solved with wget download, unable to make a proper iso using just powershell, using a mkisofs windows port

is this of use ?

### DIFF
--- a/contrib/create-basic-configdrive.ps1
+++ b/contrib/create-basic-configdrive.ps1
@@ -6,14 +6,15 @@ This tool creates a basic config-drive ISO image.
 .DESCRIPTION
 Usage: .\create-basic-configdrive.ps1 -H HOSTNAME -S SSH_FILE [-d|-e|-i|-n|-t|-h]
 Options:
-    -d URL      Full URL path to discovery endpoint.
-    -e IP:PORT  Address for client communication.
-    -H HOSTNAME Machine hostname.
-    -i IP:PORT  Address for server communication.
-    -n NAME     etcd node name.
-    -S FILE     SSH keys file.
-    -t TOKEN    Token ID from https://discovery.etcd.io.
-    -h          This help
+    -d URL             Full URL path to discovery endpoint.
+    -e http://IP:PORT  Adviertise URL for client communication.
+    -H HOSTNAME        Machine hostname.
+    -i http://IP:PORT  URL for server communication.
+    -l http://IP:PORT  Listen URL for client communication.
+    -u http://IP:PORT  Listen URL for server communication.
+    -n NAME            etcd node name.
+    -S FILE            SSH keys file.
+    -t TOKEN           Token ID from https://discovery.etcd.io.
 
 #>
 
--- a/contrib/create-basic-configdrive.ps1
+++ b/contrib/create-basic-configdrive.ps1
@@ -105,7 +105,7 @@ if ( -not ((Get-Content $SSH_FILE) -match $REGEX_SSH_FILE)) {
 }
 
 if (($TOKEN)) {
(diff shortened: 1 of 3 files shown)


crawford:
This isn't the appropriate repo. This is for building and publishing CoreOS.

https://github.com/coreos/docs would be more appropriate.

fivethreeo:
Not to be difficult but, the other scripts in this directory does the same thing, only on linux.

And it is a contrib dir, so why is this contrib directory here at all?

Is it going away and just some legacy stuff?

Docs seem to do docs only, so where should scripts go?

Just asking since I dont get why.

crawford:
I had a more detailed response but my browser spazzed and threw it away...

These scripts aren't intended to be used by end users. These are what we use to actually build and publish CoreOS images. They are all bash scripts because we use Linux exclusively in our build pipeline.

Your scripts are geared toward a user creating an instance from a published image. They would be better suited in the documentation.

crawford:
My apologies, you are right (thanks to @marineam for pointing it out). I was not familiar with the contrib directory.

fivethreeo:
No problem :)

Just wondering if anyone has any suggestions on this?

Esp. about how to do gpg validation and try to remove the need for ported linux tools.
 
Gpg needs some installer to work so cannot work "automagically".

I tried two different solutions for making the right iso using  some .net com.

Also downloading works kind of badly for some s3 downloads, also powershell is not a must, if there are other automation tools on windows better suited I take suggestions. :)

marineam:
In time we will have some Go based tools to help with installing CoreOS and
they are at least more likely to be cross platform but there isn't an ETA
on that right now.

fivethreeo:
Anyone working on this? If there is any code i'll gladly take a look at it.

marineam:
Sadly no, nothing that is directly applicable to your use case has been started yet. So for now these scripts will have to do. Could you squash your commits please? Seems sad to disable certificate verification in wget but I presume that's just because it isn't able to use Windows' own certificate store right? Or if that isn't the case why is it disabled?

fivethreeo:
Will squash them. I just turned it off because it did not work without it,
will investigate.

On 4 September 2015 at 21:14, Michael Marineau <notifications@github.com>
wrote:

> Sadly no, nothing that is directly applicable to your use case has been
> started yet. So for now these scripts will have to do. Could you squash
> your commits please? Seems sad to disable certificate verification in wget
> but I presume that's just because it isn't able to use Windows' own
> certificate store right? Or if that isn't the case why is it disabled?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/coreos/scripts/pull/452#issuecomment-137828677>.
>



-- 
Øyvind Saltvik

fivethreeo:
Could use --ca-certificate, but then it needs to download a trusted certificate first then we are sort of at square one again ;)

marineam:
Yeah, if there isn't already a root ca store wget can use it isn't worth worrying about right now.

fivethreeo:
I think you can just close this pr.

I got something better to **go** for. :)

https://github.com/fivethreeo/create-basic-configdrive/

fivethreeo:
Got the configdrive in go working in windows, create vdi will be next. Closing pull.

fivethreeo:
The tools at https://github.com/coreos/scripts/tree/master/contrib

Ported (and working) to golang can be found here:

https://github.com/fivethreeo/create-basic-configdrive/
https://github.com/fivethreeo/create-coreos-vdi/

PR #453, opened 2015-09-04
init commit of interoute oem build

xied75:
Dear Team, I believe this PR plus https://github.com/coreos/coreos-overlay/pull/1511 should pretty much get things going?

Best,

Dong


marineam:
Since it isn't obvious from the PR, what is the difference between this ovf template and our existing template?

xied75:
Our vmx version is 8, the existing one is 7; also we use e1000 and scsi etc. I thought to reuse the existing one and do some sed replacement, but then I feel it's equally mess. Besides it's a bit risky to depends on something 'external' in that someone else might later modify the stock ovf file and break our image, given our hypervisor could be 'stable' over a long time. :)

marineam:
LGTM, will merge after the overlay

PR #455, opened 2015-09-06
create-basic-configdrive: use genisoimage (replaces mkisofs on jessie), force user to specify token/discovery url

jeremy-clerc:
Hello,

I have discovered two problem while going through: https://coreos.com/os/docs/latest/booting-on-virtualbox.html

At some point I have to use create-basic-configdrive

1. I am running debian jessie, mkisofs has been replaced by genisoimage
2. The doc should be updated but to avoid errors, if no token or discovery url is specified, it should fail to generate the iso, else etcd will fail to start.

This pull request fixes these problem.

Thanks

### DIFF
--- a/contrib/create-basic-configdrive
+++ b/contrib/create-basic-configdrive
@@ -69,8 +69,12 @@ if [ $(id -u) -eq 0 ]; then
 fi
 
 # mkisofs tool check
-if ! which mkisofs &>/dev/null; then
-    echo "$0: mkisofs tool is required to create image." >&2
+if which mkisofs &>/dev/null; then
+    MKISOFS="mkisofs"
+elif which genisoimage &>/dev/null; then
+    MKISOFS="genisoimage"
+else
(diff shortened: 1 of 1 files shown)


contrib/create-basic-configdrive
```diff
@@ -115,10 +119,9 @@ fi
 
 if [ ! -z "$TOKEN" ]; then
     ETCD_DISCOVERY="${DEFAULT_ETCD_DISCOVERY//TOKEN/$TOKEN}"
-fi
-
-if [ -z "$ETCD_DISCOVERY" ]; then
-    ETCD_DISCOVERY=$DEFAULT_ETCD_DISCOVERY
+elif [ -z "$ETCD_DISCOVERY" ]; then
+    echo "$0: You need to specify a discovery token or a discovery URL." >&2
+    exit 1
```
mischief: it is not required, so please remove this

mischief:
really the script should be fixed to not include a discovery token if none was given.

PR #456, opened 2015-09-06
catalyst: re-enable ccache

marineam:
Now that ccache is turned on by default in the profile portage complains
a lot if ccache isn't actually installed, sleeping 5 seconds for each
error message. Since pkgcache is in use ccache isn't going to make that
much of a difference but getting rid of those 5 second sleeps will. :)


mischief:
lgtm

PR #458, opened 2015-09-11
Support for arm64 target ramdisk and arm64 grub modules 

andrejro:


build_packages
```diff
@@ -222,9 +222,18 @@ break_dep_loop sys-apps/util-linux udev,systemd sys-apps/systemd cryptsetup
 break_dep_loop sys-apps/systemd cryptsetup
 
 info "Merging board packages now"
+setup_qemu_static "${BOARD_ROOT}"
 sudo -E "${EMERGE_CMD[@]}" "${EMERGE_FLAGS[@]}" \
     @system coreos-devel/board-packages
 
+# update bootengine.cpio and force rebuilt the initramfs
```
glevand: Please put the full comment from the patch here. It tells why we are doing this.

common.sh
```diff
@@ -992,3 +992,17 @@ BOAT
...
+    sudo cp /usr/bin/qemu-aarch64 "${root_fs_dir}"/usr/bin/qemu-aarch64-static
+  fi
+}
+
+clean_qemu_static() {
+  local root_fs_dir="$1"
+  if [[ ${BOARD} == "arm64-usr" && -f "${root_fs_dir}/usr/bin/qemu-aarch64-static" ]]; then
+    sudo rm "${root_fs_dir}"/usr/bin/qemu-aarch64-static
+  fi
+}
```
glevand: In general I think having these available could be useful.

setup_board
```diff
@@ -155,6 +155,12 @@ fi
 if [ "${FLAGS_skip_toolchain_update}" -eq "${FLAGS_TRUE}" ]; then
   UPDATE_ARGS+=" --skip_toolchain_update"
 fi
+if [[ "${BOARD}" == "arm64-usr" ]]; then
+  [[ ! -d /etc/portage/package.use ]] && sudo mkdir /etc/portage/package.use
+  sudo_clobber "/etc/portage/package.use/grub" <<EOF
+sys-boot/grub grub_platforms_arm64
+EOF
+fi
```
glevand: I think it would be OK (and simpler) to just build the arm64 modules unconditionally by adding arm64 to GRUB_PLATFORMS in coreos-overlay/profiles/coreos/base/make.defaults.

Would there be any problem in doing it that way?

common.sh
```diff
@@ -992,3 +992,17 @@ BOAT
   echo -e "${V_VIDOFF}"
   die "$* failed"
 }
+
+setup_qemu_static() {
+  local root_fs_dir="$1"
+  if [[ ${BOARD} == "arm64-usr" && -f "${root_fs_dir}/sbin/ldconfig" ]]; then
```
glevand: Would calling this without an ldconfig be an error?  Maybe have a message?

setup_board
andrejro: Not calling clean_qemu_static leaves the option to chroot in the target
directory, which could be useful but it is not necessary.

common.sh
andrejro: Checking for "/sbin/ldconfig" just ensures that a basic system is already present
in the target directory.

build_library/grub_install.sh
```diff
@@ -192,6 +196,26 @@ case "${FLAGS_target}" in
         sudo cp "${BUILD_LIBRARY_DIR}/menu.lst" \
             "${ESP_DIR}/boot/grub/menu.lst"
         ;;
+    arm64-efi)
+        info "Installing default arm64 UEFI bootloader."
+        sudo mkdir -p "${ESP_DIR}/EFI/boot"
+        #FIXME(andrejro): shim not ported to aarch64
+        # Use the test keys for signing unofficial builds
+#        if [[ ${COREOS_OFFICIAL:-0} -ne 1 ]]; then
```
marineam: Just leave out the unused code, it can easily be pasted in later. :) Leaving the FIXME is fine though

setup_board
marineam: @glevand Well, right now building the arm64 grub target requires the arm64 cross compiler which isn't in the SDK until setup_board is used. It would be nice to avoid changing grub use flags but we'll need to adjust how we bootstrap the SDK first.

One slightly different scheme that may feel less awkard is split the grub ebuild into two, with a special grub-arm64 ebuild that can depend on the cross compiler. Then no use flag twiddling is required. Off the top of my head I'm not sure which I favor. Perhaps go with this for now but it can be tweaked later.

common.sh
marineam: Should these functions just fail in that case instead of noop?

build_packages
```diff
@@ -222,9 +222,21 @@ break_dep_loop sys-apps/util-linux udev,systemd sys-apps/systemd cryptsetup
...
 
 info "Merging board packages now"
+setup_qemu_static "${BOARD_ROOT}"
 sudo -E "${EMERGE_CMD[@]}" "${EMERGE_FLAGS[@]}" \
     @system coreos-devel/board-packages
 
+# Update bootengine.cpio and force rebuilt the initramfs:
+# as the ldconfig as called in the "/usr/sbin/update-bootengine"
+# script does not work for arm64 architecture, re-run it manually
+# and rebuild the kernel initramfs.
```
marineam: Is there a fix we can make to the ebuild or update-bootengine script so it works the first time instead of requiring this step?

marineam: Also, if it didn't work the first time wouldn't the build have already failed due to the kernel ebuild failing when it tries to use bootengine.cpio?

build_library/grub_install.sh
andrejro: Ok, removed.

build_packages
andrejro: I modified this part, now should be cleaner.
Please consider also the two pull requests:
https://github.com/coreos/bootengine/pull/70
https://github.com/coreos/coreos-overlay/pull/1536

marineam:
LGTM but before merging could you squash some of the new commits? Don't need the addcode->removecode steps in the final history :)

common.sh
andrejro: Fixed.

setup_board
andrejro: I removed this commit. It is not clear to me how do you want trigger the grub-arm64 installation.
Meanwhile, just run:
USE="grub_platforms_arm64" sudo emerge grub
before calling build_image for the arm64 target.

marineam:
LGTM!

setup_board
glevand: I think it is too much trouble to run emerge by hand, so I added this back in with PR https://github.com/coreos/scripts/pull/478.

PR #460, opened 2015-09-16
ROADMAP: remove strudel

philips:
We have de-emphasized strudel and are concentrating on testing/release/ignition for now. Remove it for the roadmap.

cc @pbx0


marineam:
I generally drop things to the unplanned section rather than deleting.

pbx0:
SGTM either way, philips suggested it may come back at some point but with different requirements/goals.

marineam:
K, if a return would mean a larger redesign might as well drop it entirely for now. so LGTM

PR #461, opened 2015-09-17
Add arm64 qemu image support

andrejro:
Enables the creation of the qemu launch script for arm64 images.


marineam:
LGTM, I think we can simplify this down a bit but that can happen later. For example the newer style of device configuration used for the disk and nic should be equally applicable to x86_64 so that need not be conditional. (or at least I think so, haven't tried it)

PR #462, opened 2015-09-21
grub: move standard kernel options to grub

marineam:
In order to boot properly we need `rootflags=rw mount.usrflags=ro` on
the command line. These have been build into the kernel directly but for
arm64 builds the built in options seem to be ignored.

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -63,35 +63,29 @@ if [ -z "$linux_console" ]; then
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


marineam:
Just to follow up, looks like to get arm64 working with built-in options we'd just need to toggle another option, it doesn't combine built-in with bootloader provided options by default like x86 does. *but* I'm voting we just drop the built-in stuff and just use grub so all of this is in one place.

marineam:
Oops, mistakenly merged the kernel change before this one.... merging now but comments are still welcome.

PR #463, opened 2015-09-23
For upstream hikey

andrejro:
Add devicetree support for bare-metal platforms.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -284,6 +284,12 @@ finish_image() {
   sudo cp "${root_fs_dir}/usr/boot/vmlinuz" \
        "${root_fs_dir}/boot/coreos/vmlinuz-a"
 
+  # Copy devicetree blob
+  if [[ -h "${root_fs_dir}/usr/boot/devicetree" ]]; then
+    sudo cp "${root_fs_dir}/usr/boot/devicetree" \
+         "${root_fs_dir}/boot/coreos/devicetree-a"
+  fi
+
   # Record directories installed to the state partition.
   # Explicitly ignore entries covered by existing configs.
   local tmp_ignore=$(awk '/^[dDfFL]/ {print "--ignore=" $2}' \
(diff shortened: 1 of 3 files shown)


marineam:
Hm, if the device tree is in the /usr image like this we will need entirely different images for different pieces of hardware. So a couple questions:
 - do we need to be able to upgrade the device tree? If not we can inject the right one into the ESP or OEM partition in image_to_vm or similar late-stage step.
 - if we do need to upgrade device tree can we ship all supported board trees in the initrd and detect the right one to use?

andrejro:
On Wed, Sep 23, 2015 at 09:00:41AM -0700, Michael Marineau wrote:
> Hm, if the device tree is in the /usr image like this we will need entirely different images for different pieces of hardware. So a couple questions:
>  - do we need to be able to upgrade the device tree? If not we can inject the right one into the ESP or OEM partition in image_to_vm or similar late-stage step.

The device tree should be considered atomic with the kernel. It should
be updated whenever the kernel image is updated.

Also, keep in mind that different hardware will have also different
kernels, not only device tree blobs (dtb file).

>  - if we do need to upgrade device tree can we ship all supported board trees in the initrd and detect the right one to use?

No, the device tree is needed by the kernel and has to be loaded
along with it.

> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/coreos/scripts/pull/463#issuecomment-142647985

--
Andrej Rosano                                           Inverse Path Srl
<andrej@inversepath.com>                      http://www.inversepath.com

0x01939B21            5BB8 574E 68E8 D841 E18F  D5E9 CEAD E0CF 0193 9B21

marineam:
Long term we shouldn't need different kernels for different hardware, needing the special 3.18 kernel for hikey is strictly temporary until mainline is working properly. Also according to @glevand a device tree shouldn't be needed when booting via UEFI, but if that isn't quite working right yet we can deal with the dtb as a temporary measure.

Given that both the 3.18 kernel and the dtb are temporary and they need to be upgraded together is there any way to just embed the dtb in the kernel image so there is less to clean up when mainline and UEFI are all working?

crawford:
Closing due to inactivity.

PR #467, opened 2015-10-27
Fixed typo on URI for DEFAULT_ETCD_DISCOVERY.

dolfdijkstra:
The DEFAULT_ETCD_DISCOVERY parameter had a typo in the scheme part of the URI. It missed a colon after https.
This lead to failure on the discovery of the cluster when the token based discovery was used as the https://discovery.etcd.io could not be reached during the boot/configuration process.


mischief:
lgtm, thank you

PR #470, opened 2015-11-06
grub: add and set coreos.randomize_disk_guid kernel param

pbx0:
coreos.first_boot=1 will no longer trigger disk-guid randomization, so
manual ignition triggers in diskless/pxe scenarios may succeed.  Instead
we explicitly request the randomization when first_boot=1 was added by
grub finding the 00000000-0000-0000-0000-000000000001 disk-guid.


marineam:
LGTM

PR #473, opened 2015-11-10
build_library: directly create streamOptimized vmdk with qemu-img

vcaputo:
Simple optimization and eliminates dependency on vmdk-convert, which
isn't the greatest quality code nor is it receiving much community
attention.

May fix https://github.com/coreos/bugs/issues/868

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -480,12 +480,7 @@ _write_vmdk_scsi_disk() {
 }
 
 _write_vmdk_stream_disk() {
-    # requires two pass conversion, qemu-img doesn't properly support the
-    # stream-optimized VMDK format. The special vmdk-convert tool only takes
-    # VMDK images as an import format.
-    local tmpvmdk="${VM_TMP_DIR}/tmp.vmdk"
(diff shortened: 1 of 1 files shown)


mischief:
any chance this can be tested on vmware?

crawford:
Yep, we have VMware licenses. We'll also need to test the other consumers of these images. I'll follow up offline.

marineam:
A big-ass patch that was reportedly needed to make stream vmdk images work never landed in QEMU. Related old PR: https://github.com/coreos/coreos-overlay/pull/1009

I have no idea if any of the smaller tweaks that have landed since then are sufficient to resolve whatever issues were going on. VMware contributed the vmdk-convert tool specifically because it wasn't possible to create working stream vmdks with QEMU.

marineam:
In short I don't know if this fixes coreos/bugs#868 but is very likely to create VMDK images that are even more broken than they currently are.

marineam:
There may be something entirely different going on like lbzip2 doing something weird, the bug report noted that building CoreOS locally was working but our release images were not.

vcaputo:
Closing this until we have a means of testing the stream optimized vmdk and hopefully reproducing the reported errors.

PR #475, opened 2015-11-11
grub: disable console blanking

vcaputo:
The console often contains very useful information in the event of a
hard crash, in such situations there's no ability to unblank the console
via keypress because the kernel won't handle the interrupt.

Since CoreOS is a server/cluster operating system, there won't generally
be monitors connected benefitting from a blanked console.  Disabling the
blanking altogether allows the frame buffer contents to always be
visible, even when the kernel can't handle keypresses.

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -77,7 +77,7 @@ else
 fi
 
 # Assemble the options applicable to all the kernels below
-set linux_cmdline="rootflags=rw mount.usrflags=ro $linux_root $linux_console $first_boot $oem_id $linux_append"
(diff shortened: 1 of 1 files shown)


mischief:
lgtm

vcaputo:
@marineam any thoughts?

marineam:
Is it just the blank-on-getty-start you are concerned about? We could also control that with a drop-in the getty service itself iirc

vcaputo:
what I'm aiming for is preventing a vga framebuffer containing potentially valuable crash information from having its output off when the crash cart is plugged in

marineam:
ahhhh, ok. sorry, I misread things. Yes that sounds fine.

LGTM

PR #477, opened 2015-11-19
common: don't get env list from chromite

marineam:
This call to a python script is merely used to get a static list of
environment variables that happens to be defined in that repo.


mischief:
lgtm

PR #479, opened 2015-11-20
Add support for Parallels Desktop VM images

bassamtabbara:
The pull request add support for Parallels Desktop VM images. Two new image types have been added:

- parallels - this produces VM images with extension pvm.tgz that can be loaded directly into Parallels Desktop
- vagrant_parallels - this produces a Vagrant box that works with parallels vagrant provider (http://parallels.github.io/vagrant-parallels/)

Just like vmdk and others we rely on qemu-img to convert raw images. Support for Parallels disk images was added to qemu-img in version 2.4 and requires the following related PR coreos/portage-stable#368.

I had to do a slight refactoring to add a concept of a VM_LAYOUT to support Parallels. Just like a disk layout, a vm layout defines how the VM is organized. Parallels Desktop requires a specific directory structure for its configuration and also for the disks. It not as simple as vmware vmx + vmdk side-by-side.

I also removed the box files from the actual image since there are not needed in /usr/share/oem.

Finally, oem-vagrant has been updated in coreos/coreos-overlay#1653 to suppress warnings regarding guest tools.

### DIFF
--- a/build_library/parallels_pvm.sh
+++ b/build_library/parallels_pvm.sh
@@ -1,0 +1,621 @@
+#!/bin/bash
+
+# Mostly this just copies the below XML, but inserting random MAC address
+# and UUID strings, and other options as appropriate.
+
+SCRIPT_ROOT=$(readlink -f $(dirname "$0")/..)
+. "${SCRIPT_ROOT}/common.sh" || exit 1
+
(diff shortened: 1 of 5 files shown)


bassamtabbara:
Replaced with #599

PR #481, opened 2015-11-24
prune_images: keep newer-than-latest builds

marineam:
This may include recent failures or builds derrived from latest, such as
those made by image_set_group.


mischief:
lgtm

PR #482, opened 2015-11-25
workaround for bootstrap_sdk on an Ubuntu host

bassamtabbara:
workaround for bootstrap_sdk on an Ubuntu host where /dev/shm is a symlink to /run/shm. Since we mount the hosts /dev (for losetup) this interferes with building python 2.7. The workaround is to disable the /dev/shm during python builds. A longer term fix would be to not mount the hosts /dev. Thanks for @marineam for suggesting the fix on IRC


crawford:
Wow, sorry for letting this sit for so long. Can you clean up the commit message per https://github.com/coreos/scripts/blob/master/CONTRIBUTING.md? Otherwise, LGTM.

crawford:
Superseded by https://github.com/coreos/scripts/pull/482.

PR #483, opened 2015-12-01
Updates for jenkins builds

marineam:
- Support arbitrary $GNUPGHOME values, avoiding conflicts in $HOME for jenkins builders.
 - Allow setting COREOS_BUILD_ID in the environment so it can be mapped directly to a jenkins job instead of using the date.


vcaputo:
lgtm

PR #484, opened 2015-12-01
sdk_lib: cleanup to support non-standard environs

crawford:


sdk_lib/enter_chroot.sh
```diff
@@ -185,9 +185,9 @@ generate_locales() {
     # with long multibyte strings.  Newer setups have this fixed,
     # but locale-gen doesn't need to be run in any locale in the
     # first place, so just go with C to keep it fast.
-    PATH="/usr/sbin:/usr/bin:/sbin:/bin" LC_ALL=C \
-      chroot "${FLAGS_chroot}" locale-gen -q -u \
-      -G "$(printf '%s\n' "${gen_locales[@]}")"
+    chroot "${FLAGS_chroot}" /usr/bin/env \
```
vcaputo: The stuff we execute under the chroot shouldn't need any changes WRT the host environment, just changing the shebang lines of these scripts which run on the host to include the /usr/bin/env should be sufficient, no?

crawford: The path that the chroot needs is very different than the path of my host. Before, we could override the host path and use that in the chroot because it was compatible with our systems. This change just ensures that the host and chroot paths stay separate (which seems more correct).

crawford: I could drop the call to /usr/bin/env though.

marineam:
LGTM

vcaputo:
lgtm

sdk_lib/enter_chroot.sh
crawford: No, I can't.

PR #485, opened 2015-12-01
common: set properly COREOS_SDK_VERSION

marineam:
Broken in 0b6acf86 when we started selectively reading version.txt


crawford:
LGTM

PR #486, opened 2015-12-01
enter_chroot: always bind $GNUPGHOME to the default path

marineam:
The path of $GNUPGHOME outside the chroot may not really make sense
inside the chroot. Although that's probably not a big deal there's no
need to keep the outside value. Instead just bind it to the usual spot.


mischief:
lgtm

PR #487, opened 2015-12-02
build_image: fix generation of version.txt

marineam:
The generation of version.txt was the only thing depending on sourcing
the deprecated BUILD, BRANCH, and PATCH values from version.txt which
common.sh no longer does since 0b6acf86. Derive them instead.


mischief:
lgtm

PR #491, opened 2015-12-16
enter_chroot: fix chroot root bind command

marineam:
Commit 09851b84 didn't do a recursive bind by mistake, so if the host
system has anything mounted under the chroot directory for some reason
the bind would hide those mounts. Recursive ensures existing mounts
remain exposed as they did before.


mischief:
lgtm

PR #494, opened 2016-01-28
release_util: use BOTO_PATH instead of searching for .boto

marineam:
This simplifies the code a little and allows BOTO_PATH or BOTO_CONFIG to
be set in the environment. Now configs may be in arbitrary locations.


mischief:
lgtm

PR #495, opened 2016-02-01
core_date: dumb little script so I can stop doing math

marineam:
Compute coreos version for a date or the date of a coreos version.
Arguments are passed to `date` except for `-v` which takes an optional
version value and converts it to a date. `-v` must be the first arg.

    $ ./core_date
    945
    $ ./core_date -d wed
    947
    $ ./core_date -v
    Fri Jan 29 00:00:00 UTC 2016
    $ ./core_date -v +%D
    01/29/16
    $ ./core_date -v 1000
    Sun Mar 27 00:00:00 UTC 2016
    $ ./core_date -v 1000 +%D
    03/27/16

### DIFF
--- a/core_date
+++ b/core_date
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Copyright (c) 2016 The CoreOS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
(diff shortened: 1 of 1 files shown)


mischief:
lgtm, but maybe some equivalent functionality should live in https://github.com/coreos/mantle/blob/master/sdk/version.go#L93 too

marineam:
at the very least that function probably will grow a time argument at some point.

PR #496, opened 2016-02-04
New build script flag --usepkgonly

marineam:
This maps to the emerge flag of the same name. Prevents building anything from source and is a quicker way to bootstrap the board for the first time.


marineam:
lots of related cleanup too :)

mischief:
besides the question, lgtm

marineam:
Yes, the shell flag library uses the `no` prefix to mean false, it is a scheme shared by many google-derived flag libraries.

PR #499, opened 2016-02-16
vm_image_util: add a small openstack image for wedging into tiny places

marineam:
Most vm images have an expanded root partiton to make them practical to
use as-is. Some deployments may not want such a large root, putting most
storage on other volumes.

/cc @rcreasey

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -11,6 +11,7 @@ VALID_IMG_TYPES=(
     pxe
     iso
     openstack
+    openstack_mini
     qemu
     qemu_uefi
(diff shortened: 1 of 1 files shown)


rcreasey:
:+1:

crawford:
This looks fine. In the future, how will we handle the combinatorial explosion of images? I don't have any particularly good ideas.

marineam:
@crawford maybe in the future fuze and the coreos-install rewrite can handle image customization on the user side of things, so we only really have to ship the base image for many systems. Full images would just be needed for targets with agents in the OEM partition and/or difficult to produce disk image formats.

PR #500, opened 2016-02-20
Support using build_image without running build_packages first

marineam:
A step towards running builds in parallel *and* efficiently, this allows build_image to download binary packages itself instead of depending on build_packages to do it.


crawford:
LGTM

PR #504, opened 2016-02-25
enable spaces in path names, add minimum error handling

knoepfchendruecker:
create-coreos-vdi fails to properly quote paths containing spaces, which may result in a failing execution of mkisofs. Due to missing error handling on this, create-coreos-vdi would also claim "Success" while the mkisofs execution actually failed.

This patch does add the missing quotes around "CONFIGDRIVE_FILE" (and around WORKDIR, just for completeness). It also adds a simple error message on a failed mkisofs operation.


glevand:
LGTM

crawford:
LGTM. Thanks.

PR #507, opened 2016-03-01
prod_image_util: move profile.env to /usr

marineam:
profile is already set up to source /usr/share/baselayout/profile.env
but it never has because I forgot to add this line during the migration
to amd64-usr images. Sure took us a while to notice that one... :(


crawford:
lol. LGTM.

mischief:
lgtm

PR #508, opened 2016-03-09
core_promote: upload gce image name to private build bucket

marineam:
The build bucket should now be writable anyone performing releases,
allowing us to ensure that the original build and public buckets are
exactly in sync.


mischief:
lgtm

PR #515, opened 2016-04-04
Add support for generating PCR configuration at image build time

mjg59:
We need to ship some PCR measurements alongside images in order to make it
easier for admins to provide an appropriate policy. Add some tooling to
generate the appropriate hashes during build, pack those into a zip file
and upload it.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -285,6 +285,9 @@ finish_image() {
   local install_grub=0
 
   local disk_img="${BUILD_DIR}/${image_name}"
+  local pcr_policy="${image_name%.bin}_pcr_policy.zip"
+  local pcr_dir="${BUILD_DIR}/pcrs"
+  mkdir -p "${pcr_dir}"
 
   # Copy kernel to support dm-verity boots
   sudo mkdir -p "${root_fs_dir}/boot/coreos"
@@ -361,6 +364,7 @@ finish_image() {
 	   "${root_fs_dir}/boot/coreos/vmlinuz-a"
   fi
 
+  ${BUILD_LIBRARY_DIR}/generate_kernel_hash.sh "${root_fs_dir}/boot/coreos/vmlinuz-a" ${COREOS_VERSION} >${pcr_dir}/kernel.config
   rm -rf "${BUILD_DIR}"/configroot
   cleanup_mounts "${root_fs_dir}"
   trap - EXIT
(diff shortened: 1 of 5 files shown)


build_library/build_image_util.sh
```diff
@@ -381,5 +386,8 @@ finish_image() {
             --target="${target}" --disk_image="${disk_img}" --noverity
       fi
     done
+    ${BUILD_LIBRARY_DIR}/generate_grub_hashes.py ${disk_img} /usr/lib/grub/ ${pcr_dir} ${COREOS_VERSION}
   fi
+  cd ${BUILD_DIR}
```
marineam: Hm, zip doesn't have an equivalent to tar's -C? I don't know if working directory is or ever will be significant to other parts of the script but probably should use pushd/popd or similar scheme just to avoid confusing anyone in the future.

mjg59: Not that I could find. I'll use pushd/popd.

build_library/generate_grub_hashes.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python
```
marineam: I'm guessing some day we will want to combine this with our grub install script or have a generalized implementation in grub itself so we can generate this directly from grub.cfg, avoiding accidental mismatches when things change. Short of that we are going to need some build time check and/or kola test to make sure we don't accidentally make a release with this script and grub.cfg conflicting.

build_library/build_image_util.sh
mjg59: Done

build_library/generate_grub_hashes.py
mjg59: Yeah. It's somewhat difficult to mechanically generate the list of valid grub configuration parameters, so I think a validation run and test is going to be the best approach here.

build_library/generate_kernel_hash.sh
```diff
@@ -0,0 +1,14 @@
...
+
+import hashlib
+import json
+import os
+import sys
+
+rootdir=sys.argv[1]
+version=sys.argv[2]
+
+path = rootdir + "/usr/boot/vmlinuz"
```
marineam: Can we point this at the ESP copy of the kernel (/boot) so we don't forget about this later when enabling verity?

mjg59: Oh good point.

mjg59: Done.

marineam:
LGTM

PR #516, opened 2016-04-05
Move PAM configuration to /usr at build time

mjg59:
A bunch of packages install PAM configuration fragments in /etc. Rather than modify
them all to install into /usr/lib, just move the entire directory at image build
time.


marineam:
Let's just do this with prod, the dev image generally remains in a state where emerge can be used directly.

mjg59:
Ok, updated to just do it in prod

marineam:
LGTM

PR #517, opened 2016-04-11
setup_board: remove bad pkg-config wrapper

marineam:
See https://github.com/coreos/coreos-overlay/pull/1883

Builds should be usring /usr/bin/$CHOST-pkg-config instead.


crawford:
LGTM

PR #522, opened 2016-04-18
build_image: Fix image type conditional

glevand:
@mjg59 I think this is what was intended.

### DIFF
--- a/build_image
+++ b/build_image
@@ -159,6 +159,7 @@ fi
 fix_mtab
 
 if [[ "${DEV_IMAGE}" -eq 1 ]]; then
+  IMAGE_BUILD_TYPE="dev"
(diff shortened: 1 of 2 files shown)


mischief:
rebase this please

marineam:
Did this slightly differently in https://github.com/coreos/scripts/issues/532 which leaves sbsign enabled. I figure it is worth while to make sure that keeps working :)

glevand:
@marineam It seems https://github.com/coreos/scripts/pull/532 does something different.  Even with that applied, we are still using PROD_IMAGE incorrectly in build_image_util.sh, which is what this PR tries to fix.

marineam:
Oh, I see. yeah, so this is wrong when building dev and prod at the same time. Seems like a refactoring or adding an argument is in order but this approach will work for now.

PR #526, opened 2016-04-27
Increase the size of /usr on devel images

mjg59:
devel images contain significantly more code and we're overflowing the
filesystem. Increase this, but only for devel images.

### DIFF
--- a/image_to_vm.sh
+++ b/image_to_vm.sh
@@ -115,6 +115,10 @@ trap vm_cleanup EXIT
 
 fix_mtab
 
+if [[ "${FLAGS_prod_image}" -eq "${FLAGS_FALSE}" && "${FLAGS_disk_layout}" == "" ]]; then
+    FLAGS_disk_layout=devel
+fi
+
 # Setup new (raw) image, possibly resizing filesystems
 setup_disk_image "${FLAGS_disk_layout}"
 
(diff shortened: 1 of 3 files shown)


marineam:
I think image_to_vm won't work without passing the disk format flag to use this new devel one. Otherwise when it goes to resize the disk it should detect thr incompatiblity and fail. Let's expand the rootfs in devel to be the same size as the 'vm' format and then default do devel for all vm types when processing a dev image.

mjg59:
Ok, reworked this to be useful in vm generation

image_to_vm.sh
```diff
@@ -54,6 +53,9 @@ DEFINE_string getbinpkgver "" \
 FLAGS "$@" || exit 1
 eval set -- "${FLAGS_ARGV}"
 
+# source after we've evaluated the command line
+. "${BUILD_LIBRARY_DIR}/vm_image_util.sh" || exit 1
```
marineam: VALID_IMG_TYPES is used above which comes from here. Maybe leave vm_image_util alone and do this:

```sh
if [[ "${FLAGS_prod_image}" -eq ${FLAGS_FALSE}" && "${FLAGS_disk_layout}" == "" ]]; then
    FLAGS_disk_layout=devel
fi
```

mjg59:
Ok, one more go

marineam:
LGTM

PR #528, opened 2016-04-28
core_roller_upload: use new update storage bucket

marineam:
The google domain name is still used for downloads, that will be
switched as soon as the appropriate certificates are set up.


crawford:
LGTM

PR #532, opened 2016-05-02
build_image: disable verity on arm64

marineam:
Our current scheme injects the verity hash into a free spot in x86
kernel images. This is a bad thing to try on other types ;-)

Smaller alternative to https://github.com/coreos/scripts/pull/521 and https://github.com/coreos/scripts/pull/522


mischief:
lgtm

PR #536, opened 2016-05-06
prod_image_util: do not check update keys on arm64

marineam:
arm64-usr doesn't have update_engine yet so this isn't valid.


mischief:
lgtm.. though hopefully this is fixed 'for real' soon.

PR #538, opened 2016-05-09
*: drop obsolete gentoo repo reference

marineam:
Instead of patching portage to support the `disabled` flag now we just
patch it to leave the `[gentoo]` section out of the default repos.conf.

Follow up to https://github.com/coreos/coreos-overlay/commit/585275b268da9f549d93c7c37744dcda3411cca2


mischief:
lgtm

PR #545, opened 2016-05-16
tag_release: support branches, add a branch tracking manifest, cleanup version

marineam:
- remove old BUILD/BRANCH/PATCH values
    
- To make branches easier to use this splits the branch manifest into two:
   build-????.xml is now only pins revisions of projects that do not have a
   corresponding branch (yet) while release.xml pins all revisions. Unlike
   before the script can now be used to tag branched releases.

See also https://github.com/coreos/manifest/pull/87


crawford:
LGTM

PR #549, opened 2016-05-24
Revert "*: drop obsolete gentoo repo reference"

marineam:
Reverts coreos/scripts#538

Not everyone has flushed older versions of portage out of their systems yet. Keep these compatibility configs for a bit longer.


mischief:
lgtm

PR #550, opened 2016-05-26
enter_chroot: do not export variables that weren't previously exported

marineam:
COREOS_BUILD_ID is set to a default value in common.sh if unset in the
environment. When entering the chroot this default value should not then
get promoted into the environment. Doing so causes catalyst to re-use
stale builds and multiple build_image runs to conflict with each other.

### DIFF
--- a/sdk_lib/enter_chroot.sh
+++ b/sdk_lib/enter_chroot.sh
@@ -394,7 +394,11 @@ CHROOT_PASSTHRU=(
 # Add the whitelisted environment variables to CHROOT_PASSTHRU.
 load_environment_whitelist
 for var in "${ENVIRONMENT_WHITELIST[@]}" ; do
-  [ "${!var+set}" = "set" ] && CHROOT_PASSTHRU+=( "${var}=${!var}" )
+  # skip empty/unset values
+  [[ "${!var+set}" == "set" ]] || continue
+  # skip values that aren't actually exported
+  [[ $(declare -p "${var}") == "declare -x ${var}="* ]] || continue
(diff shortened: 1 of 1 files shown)


sdk_lib/enter_chroot.sh
```diff
@@ -394,7 +394,11 @@ CHROOT_PASSTHRU=(
 # Add the whitelisted environment variables to CHROOT_PASSTHRU.
 load_environment_whitelist
 for var in "${ENVIRONMENT_WHITELIST[@]}" ; do
-  [ "${!var+set}" = "set" ] && CHROOT_PASSTHRU+=( "${var}=${!var}" )
+  # skip empty/unset values
+  [[ "${!var+set}" == "set" ]] || continue
+  # skip values that aren't actually exported
+  [[ $(declare -p "${var}") == "declare -x ${var}="* ]] || continue
```
crawford: Is this executed by the user's shell or the SDK's?

marineam: The user's shell. This whole script is.

crawford: That's what I was afraid of.

bash:

```
$ declare -p PWD
declare -x PWD="/home/alex"
```

zsh (my shell):

```
$ declare -p PWD         
typeset -x PWD=/home/alex
```

marineam: This script is bash, I presume you have bash on your box even if you don't use it by default.

crawford: -_- one of these days I'll pay attention to the shebang.

crawford:
LGTM

PR #553, opened 2016-06-01
set_lsb_release: use an alternate app id for arm64

marineam:
update_engine and CoreUpdate aren't quite ready, use an alternate id so
we can continue testing but not mistakenly serve amd64 updates to arm64.


crawford:
LGTM

PR #554, opened 2016-06-07
oem/azure: update script for new publishing schema

crawford:


marineam:
I mean, I guess it is ok. probably :-P

LGTM

PR #555, opened 2016-06-09
oem: inflate images into /var/tmp

crawford:
tmpfs shouldn't be used for large images like this. Inflate the archives
on disk instead.


marineam:
For reference you can also export TMPDIR=/var/tmp

marineam:
LGTM

PR #558, opened 2016-06-13
prod_image_util: move pam.d contents, not the directory

marineam:
The baselayout package now creates and installs /usr/lib/pam.d, causing
mv to fail. Move the contents instead which is happy with both versions.

### DIFF
--- a/build_library/prod_image_util.sh
+++ b/build_library/prod_image_util.sh
@@ -108,7 +108,9 @@ L+  /etc/ld.so.conf     -   -   -   -   ../usr/lib/ld.so.conf
 EOF
 
   # Move the PAM configuration into /usr
-  sudo mv ${root_fs_dir}/etc/pam.d ${root_fs_dir}/usr/lib
+  sudo mkdir -p ${root_fs_dir}/usr/lib/pam.d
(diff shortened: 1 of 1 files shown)


build_library/prod_image_util.sh
```diff
@@ -108,7 +108,8 @@ L+  /etc/ld.so.conf     -   -   -   -   ../usr/lib/ld.so.conf
 EOF
 
   # Move the PAM configuration into /usr
-  sudo mv ${root_fs_dir}/etc/pam.d ${root_fs_dir}/usr/lib
+  sudo mkdir -p ${root_fs_dir}/usr/lib/pam.d
+  sudo mv -n ${root_fs_dir}/etc/pam.d/* ${root_fs_dir}/usr/lib/pam.d/
```
crawford: Why no-clobber? Should we empty this directory before doing the move?

marineam: The directory isn't and shouldn't be empty following the new baselayout version. I tossed in -n on the off chance that something weird happened. What I really want is "fail if you would have clobbered" but there isn't any such option.

marineam: oh, I could do rmdir and that'll be that sanity check

marineam: updated

crawford:
LGTM

PR #561, opened 2016-06-22
*: Fix PIPESTATUS checks for bash 4.3

marineam:
The one-liner `[[ -z ${PIPESTATUS[*]#0} ]]` no longer works because the
expansion still includes spaces even if all the values are zero. Somehow
that didn't matter in bash 4.2 but it does mater in 4.3 to be consistent
with the general behavior of variables in [[ tests.


mischief:
lgtm

PR #565, opened 2016-07-04
Glance image API v1 must now be explicitly specified by the client

tjdett:
The OpenStack client `glance` option `--is-public true` is limited to Image API version 1, and so doesn't work with versions of the glance client where version 2 is the default.

`python-glanceclient` 1.0.0 was the first to do so, and was released Sep 2015, so most new users of this script will need this fix.


tjdett:
Contribution made in accordance with the DCO and the project licence on behalf of copyright holder "The University of Melbourne".

crawford:
LGTM. Thank you!

PR #571, opened 2016-07-14
rebuild_packages: new script to catch all ebuild modifications

marineam:
This is a bit of a hack, I would like something a little more
intelligent that checks for applicable metadata changes, not just any
old change in the ebuild text. That will require a bit more work/thought
and this should at least be sufficient to catch the current problems.

The issue comes down to how emerge's --usepkgonly mode works. Normally
KEYWORDS and (R)DEPEND metadata in ebuilds is used to figure out what to
install. If that metadata in the installed or binary packages is is out
of date things generally work out anyway. However when --usepkgonly is
enabled (such as in build_image) the ebuild metadata is ignored, using
only the installed and binary package metadata. Unfortunately it may be
out of date, allowing build_image to fail or otherwise behave
differently from build_packages. Perhaps there is a nice tool for
rebuilding such stale packages but emerge itself doesn't appear to.

This script should only be run after build_packages.

### DIFF
--- a/rebuild_packages
+++ b/rebuild_packages
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+# Copyright (c) 2016 The CoreOS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+SCRIPT_ROOT=$(dirname $(readlink -f "$0"))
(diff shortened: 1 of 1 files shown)


vcaputo:
hacky but it's relatively benign since it's something manually run and has potential utility... seems ok to me

PR #572, opened 2016-07-14
build_packages: fix @preserved-rebuild

marineam:
When the --usepkgonly flag was added in d327840ce this step got wrapped
in an inverted test. It should normally be enabled.


crawford:
LGTM

PR #573, opened 2016-07-18
oem/ami: write ami index files to storage after copy

marineam:
Since this writes to private storage there isn't any need to wait until
the images are public to do this. Now the final publish step only
changes permissions on the AMIs and nothing more.


mischief:
lgtm

vcaputo:
lgtm

PR #575, opened 2016-07-21
grub: abort boot if gptprio fails

marineam:
If the gptprio.next command fails to give us something to boot we
shouldn't try! In order to diagnose why the failure happened halt
immediately so the user can see the error message.

For https://github.com/coreos/bugs/issues/1471


mjg59:
LGTM

PR #577, opened 2016-08-01
Force HTTP/1.0 in fix_check_etag.sh

tjdett:
Curl and CloudFlare now support HTTP/2. While that's great news, it breaks `fix_check_etag.sh` because headers are lower-case in h2.

Forcing HTTP/1.0 has good backwards compatibility with Curl, no significant performance impact and resolves the inconsistency.

### DIFF
--- a/oem/generic/check_etag.sh
+++ b/oem/generic/check_etag.sh
@@ -2,14 +2,14 @@
 # check_etag.sh
 #
 # A tool for monitoring remote files and acting if the file has changed
-# This tool has been optimized for use on CoreOS but should work 
+# This tool has been optimized for use on CoreOS but should work
 # for most files where an ETag is exposed.
- 
+
 # by default we retrieve the alpha image, set to  a different value to retrieve
 # that release
 
 set -e -o pipefail
- 
+
 # Set default release to alpha if not specified
 release=${1:-"alpha"}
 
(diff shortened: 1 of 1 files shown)


tjdett:
ping @marineam

crawford:
LGTM

brianredbeard:
While this works, a potentially better answer is to be more explicit about the header returned via the regular expression `/^[Ee][Tt][Aa][Gg]:/`

tjdett:
@brianredbeard I agree that this solution *feels* better, and I wanted to do it originally. However when you look at the headers next to each other, you'll notice an additional difference:

```
HTTP/1.0    ETag: "bf9b2d5a78e166f24540b2303bffbe4e"
HTTP/2.0    etag:"bf9b2d5a78e166f24540b2303bffbe4e"
```

There's also a space missing in h2, which is a pain to fix with awk. While it's not like that problem can't be solved, it requires more than changing the regex.

I figured a three character change would be easier to code review because it reduces the range of inputs back to what they were previously. Given this is CoreOS's bag of custom scripts (which I'm appreciatively using) I wanted to go with something that was guaranteed to work in most environments.

Would you prefer: ```| grep -i '^etag' | sed -e 's/.*"\(.*\)"/\1/g'```?

marineam:
I'd say don't use a sed expression that'll do nothing if quotes don't happen to be used for some reason. So: `| grep -i '^etag:' | sed -e 's/.*: *//' -e 's/"//'`

That should be happy no matter what :)

tjdett:
@marineam I like it, but you missed the `g` on the second pattern match.

I've updated to use that solution. If we're all done with the bike shedding, can I get a merge? 😉

tjdett:
The switch to `sed` missed the issue of line endings. I've updated the patch one more time to fix that too.

PR #580, opened 2016-08-16
menuconfig: Handle shared architecture configuration

mjg59:
Handling configuration changes across multiple architectures is difficult.
Let's just generate a diff in the config and then apply that to the arch or
common config manually.


mischief:
lgtm

PR #581, opened 2016-08-23
tag_release: operate on copy of manifest.xml

crawford:
Without this, repo will sometimes empty the contents of manifest.xml.

### DIFF
--- a/tag_release
+++ b/tag_release
@@ -82,9 +82,12 @@ if [[ -L release.xml ]]; then
 fi
 
 # Generate a release manifest with all revisions pinned
-repo manifest -o release.xml -r
+tmp=$(mktemp --suffix repo)
+trap "rm --force ${tmp}" EXIT
+repo manifest -o ${tmp} -r
 # double check because repo may not exit with non-zero on failure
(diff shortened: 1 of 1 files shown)


vcaputo:
seems reasonable, but I wonder if we can just fix repo to not do that?

tag_release
```diff
@@ -82,9 +82,12 @@ if [[ -L release.xml ]]; then
 fi
 
 # Generate a release manifest with all revisions pinned
-repo manifest -o release.xml -r
+tmp=$(tempfile --prefix repo)
```
marineam: use mktemp

crawford: Done. Why though?

marineam: From `tempfile(1)`: `tempfile creates a temporary file in a safe manner.  It uses tempnam(3)` and later `tempfile is deprecated; you should use mktemp(1) instead.`

And from `tempnam(3)`: `Never use this function.  Use mkstemp(3) or tmpfile(3) instead.`

Yay legacy!

marineam:
LGTM

PR #582, opened 2016-08-23
catalyst: workaround for bootstrap_sdk on Ubuntu

crawford:
workaround for bootstrap_sdk on an Ubuntu host where /dev/shm is a
symlink to /run/shm. Since we mount the hosts /dev (for losetup) this
interferes with building python 2.7. The workaround is to disable the
/dev/shm during python builds. A longer term fix would be to not mount
the hosts /dev. Thanks for marineam for suggesting the fix on IRC.


vcaputo:
lgtm

PR #585, opened 2016-09-14
grub: rename oem_id variable

crawford:
On HVM instances, tty0 is actually a graphical interface. Since the
interface exposed is non-interactive, it's not possible to scroll
through the logs, which results in lots of missing information. Setting
the console to ttyS0, on the other hand, sends the output to the system
log.

PV instances don't have this issue since they are xen-based and use
hvc0.

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -72,8 +72,9 @@ if [ -n "$first_boot" ]; then
     set first_boot="coreos.first_boot=1 coreos.randomize_disk_guid=00000000-0000-0000-0000-000000000001"
 fi
 
+set oem=""
 if [ -n "$oem_id" ]; then
-    set oem_id="coreos.oem.id=$oem_id"
+    set oem="coreos.oem.id=$oem_id"
 fi
 
 # If no specific console has been set by the OEM then select based on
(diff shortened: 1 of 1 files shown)


vcaputo:
```
You can specify multiple console= options on the kernel command line.
Output will appear on all of them. The last device will be used when
you open /dev/console. So, for example:

        console=ttyS1,9600 console=tty0

defines that opening /dev/console will get you the current foreground
virtual console, and kernel messages will appear on both the VGA
console and the 2nd serial port (ttyS1 or COM2) at 9600 baud.
```

Ok, so we're extricating tty0 from /dev/console on ec2 with this change, lgtm

crawford:
Works nicely.

```
systemd[1]: Failed to start Ignition (disks).
ignition[244]: error at line 2, column 2
    1: {
    2: h
       ^
invalid character 'h' looking for beginning of object key string
```

build_library/grub.cfg
```diff
@@ -72,15 +72,20 @@ if [ -n "$first_boot" ]; then
...
+    set oem="coreos.oem.id=$oem_id"
 fi
 
 # If no specific console has been set by the OEM then select based on
 # platform, most systems use vga text as primary and ttyS0 as secondary.
 if [ -z "$linux_console" ]; then
     if [ "$grub_platform" = pc ]; then
-        set linux_console="console=ttyS0,115200n8 console=tty0"
+        if [ "$oem_id" = ec2 ]; then
+            set linux_console="console=ttyS0,115200n8"
```
marineam: Why are you not just setting `set linux_console="console=ttyS0,115200n8 console=tty0"` in the oem grub.cfg?

marineam: er, set linux_console="console=ttyS0,115200n8" that is

marineam: If you want to differentiate between hvm and pv you can use "$grub_platform" in the oem grub.cfg, it will be either "pc" or "xen" respectively

crawford: I guess we could do that. Would you prefer that solution?

crawford:
Trimmed the EC2-specific changes.

marineam:
LGTM

PR #592, opened 2016-10-05
build_library: bump vmware to hardware version 9

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1583.

mischief:
does https://github.com/coreos/scripts/blob/master/build_library/vm_image_util.sh#L732 need updating too?

crawford:
@mischief probably. good catch.

marineam:
There are a lot of differences between template_vmware.ovf and template_interoute.ovf, where the latter was based on hardware version 8. Should we pass this through a vmware tool to perform the update in case there are useful/important differences beyond just the hardware rev number?

crawford:
Yeah, I was using the workstation upgrade tool to run the migration (changes haven't been posted). I'm holding off on this for now since I'm having trouble reproducing the original failure in the bug report.

marineam:
So I guess the bug isn't applicable to vmware worksation but some other product. Considering vmware admitted to not testing w/ the older hardware versions with newer kernels we should probably go ahead and update the version anyway.

PR #593, opened 2016-10-07
Add OEM ACI building support

dm0-:
This adds a function to the `image_to_vm` library that builds an ACI file and installs it into `/usr/share/oem`.  The "building" part is a new library file that reuses the `create_prod_image` function, run in a slightly more controlled environment.

This is for coreos/bugs#1468.

### DIFF
--- a/build_library/oem_aci_util.sh
+++ b/build_library/oem_aci_util.sh
@@ -1,0 +1,124 @@
+# Copyright (c) 2016 The CoreOS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Expects BOARD, BUILD_DIR, BUILD_LIBRARY_DIR, and COREOS_VERSION in env.
+
+# There must be a manifest template included with the ebuild at
+# files/manifest.in, which will have some variable values substituted before
+# being written into place for the ACI.  Optionally, a shell script can also be
+# included at files/manglefs.sh to be run after all packages are installed.  It
+# is intended to be used to make modifications to the file system layout and
+# program paths that some included agent software might expect.
+
+# Copied from create_prod_image()
+create_oem_aci_image() {
+    local image_name="$1"
+    local disk_layout="$2"
+    local update_group="$3"
+    local base_pkg="${4?No base package was specified}"
+
+    info "Building OEM ACI staging image ${image_name}"
+    local root_fs_dir="${BUILD_DIR}/rootfs"
+    local image_contents="${image_name%.bin}_contents.txt"
+    local image_packages="${image_name%.bin}_packages.txt"
+    local image_licenses="${image_name%.bin}_licenses.txt"
+
+    start_image \
+        "${image_name}" "${disk_layout}" "${root_fs_dir}" "${update_group}"
+
+    # Install minimal GCC (libs only) and then everything else
+    set_image_profile oem-aci
+    extract_prod_gcc "${root_fs_dir}"
+    emerge_to_image "${root_fs_dir}" "${base_pkg}"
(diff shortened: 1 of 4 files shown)


build_library/oem_aci_util.sh
```diff
@@ -0,0 +1,80 @@
...
+        die_notrace "The \"${base_pkg}\" ebuild is not a meta package"
+
+    # Disable some features, and run a regular prod image build for this OEM.
+    # Note that we later depend on this function using "${BUILD_DIR}/rootfs".
+    local save_COREOS_OFFICIAL=$COREOS_OFFICIAL
+    local save_FLAGS_upload=$FLAGS_upload
+    COREOS_OFFICIAL=0
+    FLAGS_upload=1
+    mkdir -p "${BUILD_DIR}"/configroot/etc/portage/
+    echo 'dev-lang/python' >> "${BUILD_DIR}"/configroot/etc/portage/package.unmask
```
marineam: I suppose we need to use the 'dev' sub-profile or create a new one?

build_library/oem_aci_util.sh
```diff
@@ -0,0 +1,80 @@
...
+    [ -n "${ebuild}" ] || die_notrace "No ebuild exists for OEM \"${oem}\""
+    grep -Fqs '(meta package)' "${ebuild}" ||
+        die_notrace "The \"${base_pkg}\" ebuild is not a meta package"
+
+    # Disable some features, and run a regular prod image build for this OEM.
+    # Note that we later depend on this function using "${BUILD_DIR}/rootfs".
+    local save_COREOS_OFFICIAL=$COREOS_OFFICIAL
+    local save_FLAGS_upload=$FLAGS_upload
+    COREOS_OFFICIAL=0
+    FLAGS_upload=1
```
marineam: Ick, if re-using the prod image code perhaps it should be broken down into the reusable parts we want. Also why are you forcing upload?

build_library/vm_image_util.sh
```diff
@@ -444,6 +445,25 @@ install_oem_package() {
...
 }
 
+# Write the OEM ACI file into the OEM partition.
+install_oem_aci() {
+    local oem_aci=$(_get_vm_opt OEM_ACI)
+
+    [ -n "${oem_aci}" ] || return 0
+
+    local save_BUILD_DIR=$BUILD_DIR
+    BUILD_DIR=$FLAGS_to
```
marineam: Mixing state like this scares me, oem_aci_create should called as a different script.

build_library/oem_aci_util.sh
dm0-: I'll duplicate the function without those official update and upload commands, then.  And that disables uploading ... `1` is the "false" value for some reason.

build_library/oem_aci_util.sh
dm0-: I don't think we should be using `dev`, as it would leave so much unused bloat in the container (static libraries, build tools, etc.).  If unmasking a package is unacceptable here, we should have another sub-profile that doesn't mask packages.

build_library/vm_image_util.sh
dm0-: I'm not sure how calling it as a different script would make a differnece.  The functions called in `create_prod_image` rely on `BUILD_DIR` as a result of normally being used by the `build_image` script, so that still needs to be defined.  It doesn't seem to be defined properly by the `image_to_vm` script.

dm0-:
@marineam I've added the `oem-aci` subprofile to coreos/coreos-overlay#2224 and duplicated the `prod_create_image` function to remove the two variables.

dm0-:
@marineam I've added a script comparable to `build_image`.  It is called automatically when an ACI is required and missing.

PR #598, opened 2016-10-11
glance_load: Add board argument

glevand:
Add a board argument to allow arm64 downloads.

/cc @brianredbeard

### DIFF
--- a/oem/openstack/glance_load.sh
+++ b/oem/openstack/glance_load.sh
@@ -7,24 +7,48 @@
 #
 # You will need to source your glance credentials before using this script
 #
-# By default we retrieve the alpha image, set to  a different value to retrieve
-# that release
+# By default we retrieve the amd64-usr alpha image, set different values to
+# retrieve a specific release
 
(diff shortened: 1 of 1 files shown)


oem/openstack/glance_load.sh
```diff
@@ -7,24 +7,47 @@
...
 set -e -o pipefail
 
-release=${1:-"alpha"}
-if [[ "$1" != http* ]]; then
-  baseurl="https://${release}.release.core-os.net/amd64-usr/current"
-else
-  # for this convoluded trick, we take an arbitrary URL, chop it up, and try
-  # to turn it into usable input for the rest of the script.
+board="amd64-usr"
+release="alpha"
```
brianredbeard: Defining the variable in this way removes the ability to overload it it from outside of the script at runtime.  Currently merging this would break all users who simply run the script with "release=beta ./glance-load.sh"

glevand:
Allow release and board variables to be set by environment.

PR #599, opened 2016-10-17
add support for Parallels Desktop VM images

bassamtabbara:
Two new image types have been added:

1. parallels - this produces VM images with extension pvm.tgz that can be loaded directly into Parallels Desktop
2. vagrant_parallels - this produces a Vagrant box that works with parallels vagrant provider (http://parallels.github.io/vagrant-parallels/)

Just like vmdk and others we rely on qemu-img to convert raw images. Support for Parallels disk images was added to qemu-img in version 2.4.

I had to do a slight refactoring to add a concept of a VM_LAYOUT to support Parallels. Just like a disk layout, a vm layout defines how the VM is organized. Parallels Desktop requires a specific directory structure for its configuration and also for the disks. It not as simple as vmware vmx + vmdk side-by-side.

I also removed the box files from the actual image since there are not needed in /usr/share/oem.

### DIFF
--- a/build_library/write_hdd.sh
+++ b/build_library/write_hdd.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+# Generate a parallels compatible disk with XML format. qemu-img does not currently
+# support the XML disk format. If changes in the future this script can replaced simply
+# by a qemu-img convert.
+
+SCRIPT_ROOT=$(readlink -f $(dirname "$0")/..)
+. "${SCRIPT_ROOT}/common.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/build_image_util.sh" || exit 1
+
+DEFINE_string input_disk_image "" "Disk image to convert from, required."
+DEFINE_string input_disk_format "raw" "Disk image format."
+DEFINE_string output_disk "" "Path to the output disk, required."
+
+# Parse command line
+FLAGS "$@" || exit 1
+eval set -- "${FLAGS_ARGV}"
+
+# Die on any errors.
+switch_to_strict_mode
+
+if [[ ! -e "${FLAGS_input_disk_image}" ]]; then
+    echo "No such disk image '${FLAGS_input_disk_image}'" >&2
+    exit 1
+fi
+
+rm -fr "${FLAGS_output_disk}"
+mkdir -p "${FLAGS_output_disk}"
+disk_name=$(basename "${FLAGS_output_disk}")
+
+touch "${FLAGS_output_disk}"/"${disk_name}"
+
+# this id is a constant and identifies the first snapshot of a disk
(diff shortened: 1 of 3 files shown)


bassamtabbara:
Hi, #479 was open for almost a year and I deleted the fork. This is a replacement for it. Can you please review?

glevand:
It seems adding the vm_layout support is independent of the parallels support, so I would  say split that off into its own patch.

bassamtabbara:
@glevand the vm_layout is needed for parallels support. its the first vm format where the file layout is different from the actual disk format.

marineam:
I guess my main thought is VM_LAYOUT is adding complexity in an already pretty confusing script (and I'm the one who wrote it). Since there isn't much in the way of code to create the pvm layout what about `_write_pvm_tgz_layout` being a bundle instead and `_write_pvm_layout` just going into the box bundle code?

bassamtabbara:
@marineam the vm_layout is shared between the parallels vagrant box and parallels pvm_tgz. I could make a new bundle for each, but I really wanted to leverage the existing box BUNDLE_FORMAT as well. I see what you're saying about more confusion but this factoring reduces duplication.

marineam:
What is shared? _write_pvm_tgz_layout and _write_pvm_layout duplicate the same mkdir and two mv command. Doesn't seem like enough duplication to worry about anyway.

bassamtabbara:
@marineam I removed the vm_layout as you suggested. I agree it looks more maintainable. We're running a build and will test this out on our side, but I wanted to get this out for review.

bassamtabbara:
@marineam testing look good with these changes. Parallels support seems to work fine for both vagrant and pvm_tgz images. Let me know if you see anything else.

build_library/write_hdd.sh
```diff
@@ -0,0 +1,91 @@
...
+    exit 1
+fi
+
+rm -fr "${FLAGS_output_disk}"
+mkdir -p "${FLAGS_output_disk}"
+disk_name=$(basename "${FLAGS_output_disk}")
+
+touch "${FLAGS_output_disk}"/"${disk_name}"
+
+snapshot_uuid="{5fbaabe3-6958-40ff-92a7-860e329aab41}"
```
marineam: Is this supposed to be static? If so add a comment what it actually means, if not call uuidgen

marineam:
Just one small uuid detail, otherwise looks good

bassamtabbara:
@marineam added a comment. thanks for the review.

marineam:
Thanks! These images should be in the next alpha

marineam:
Fyi, small detail that we missed: https://github.com/coreos/scripts/pull/605

bassamtabbara:
Oops. Thanks for fixing!

PR #602, opened 2016-10-21
build_jobs: import current jenkins build scripts

marineam:


marineam:
My intent here is to start maintaining these in git instead of directly in Jenkins so it is easier to change/improve the jobs.

PR #608, opened 2016-11-02
oem/ami: make snapshots public

crawford:
This is needed so that users can copy the public AMIs into their own
accounts.


crawford:
Fixes https://github.com/coreos/bugs/issues/1090.

PR #610, opened 2016-11-15
grub.cfg: Enable arm64 ACPI

glevand:
By default, ACPI is disabled on arm64 even if support is built in to the kernel (```CONFIG_ACPI=y```).

This change adds a new grub.cnf variable ```platform_options``` and sets ```acpi=force``` to enable ACPI on arm64 platforms.

Needs https://github.com/coreos/coreos-overlay/pull/2273 (arm64_defconfig: Enable ACPI).

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -96,6 +96,11 @@ if [ -z "$linux_console" ]; then
     fi
 fi
 
+set extra_options=""
+if [ "$grub_cpu" = arm64 ]; then
+	set extra_options="acpi=force"
+fi
+
 set suf=""
 
 # UEFI uses linuxefi/initrdefi instead of linux/initrd except for arm64
@@ -106,7 +111,7 @@ if [ "$grub_platform" = efi ]; then
 fi
 
 # Assemble the options applicable to all the kernels below
-set linux_cmdline="rootflags=rw mount.usrflags=ro consoleblank=0 $linux_root $linux_console $first_boot $oem $linux_append"
+set linux_cmdline="rootflags=rw mount.usrflags=ro consoleblank=0 $linux_root $linux_console $first_boot $extra_options $oem $linux_append"
 
 # Re-implement grub_abort() since no command exposes it.
 function abort {


glevand:
Rebased to latest.

glevand:
Rebased to latest.  Renamed ```platform_options``` to ```extra_options```.

marineam:
Seems like this would be a good candidate to put into CONFIG_CMDLINE in the kernel so it can be dropped once the upstream kernel starts enabling ACPI automatically when appropriate. (I presume that will happen some day at least)

Note: this requires moving CONFIG_CMDLINE from common to arch defconfig, also iirc the arm64 kernel requires a second setting to allow CONFIG_CMDLINE to be combined with the runtime specified cmdline like x86 kernels do. Otherwise CONFIG_CMDLINE is only used if there is no runtime cmdline.

glevand:
Based on the arm64 ACPI discussions, I don't think ACPI enabled will be the default anytime soon, but we can certainly put it into the kernel config.  I'll look into it.

glevand:
@marineam On looking into the arm64 kernel's command line handling, either the built in is used, or the one passed from the bootloader is used.  There is no way to append.

See: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/drivers/of/fdt.c?h=v4.8#n1066

glevand:
Closing.
See  [Enable ACPI in arm64 CONFIG_CMDLINE](https://github.com/coreos/coreos-overlay/issues/2298).

glevand:
Reopening this PR since arm64 kernel maintainers rejected my patch to add CONFIG_CMDLINE_EXTEND support to arm64.  See:  http://www.spinics.net/lists/arm-kernel/msg546922.html

If we want to set ```acpi=force``` in our kernel config, then we would need to maintain the kernel patch that adds CONFIG_CMDLINE_EXTEND.

marineam:
Aw, damn. Sorry for the wild goose chase. I could have sworn that option already existed. GRUB config it is then.

PR #611, opened 2016-11-18
scripts: Add arm64 verity support

glevand:
Tested with amd64 coreos_production_qemu_uefi and did not see any regressions.

Depends on:
https://github.com/coreos/grub/pull/42 "Add arm64 verity support" 
coreos/coreos-overlay#2277 "coreos-sources: Add arm64 verity hash"

cc: @mjg59

### DIFF
--- a/build_image
+++ b/build_image
@@ -86,11 +86,6 @@ switch_to_strict_mode
 
 check_gsutil_opts
 
-# Inserting the verity hash into the kernel assumes x86_64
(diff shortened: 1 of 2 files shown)


glevand:
Rebased to latest.

glevand:
Uses: https://github.com/coreos/coreos-overlay/pull/2396 (sys-boot/grub: update to get arm64 verity)

PR #614, opened 2016-11-26
Delete download.sh

MichaNice:
Unauthorized


crawford:
We use this to sign images.

PR #616, opened 2016-12-05
build_jobs: update manifest-builds location

crawford:


marineam:
Is the https url usable? Or does everything need to convert to ssh?

marineam:
Nm, you changed it :)

marineam:
Changing the manifest url to ssh:// causes all github repos to be fetched over ssh:// (the urls in the XML manifest are relative to the manifest url) so please update the comments on all the jobs to note that ssh read-only access is required to all repos.

crawford:
Going to do this on the new Jenkins instead.

PR #619, opened 2016-12-20
check_root: make #!/usr/bin/env a special case

dm0-:
Make detecting issues like coreos/bugs#1139 easier.

### DIFF
--- a/build_library/check_root
+++ b/build_library/check_root
@@ -177,9 +177,22 @@ def check_shebang():
                 line = fd.readline(80)
                 if not line.startswith(b"#!"):
                     continue
-                cmd = line[2:].rstrip().split(None,1)[0]
+                args = line[2:].rstrip().split(None, 2)
+                cmd = args.pop(0)
+                if cmd in ('/usr/bin/env', '/bin/env') and args:
+                    prog = args.pop(0)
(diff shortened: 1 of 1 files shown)


build_library/check_root
```diff
@@ -177,9 +177,20 @@ def check_shebang():
                 line = fd.readline(80)
                 if not line.startswith(b"#!"):
                     continue
-                cmd = line[2:].rstrip().split(None,1)[0]
+                words = line[2:].rstrip().split(None, 1)
```
marineam: A couple things, if either env or the prog have arguments this will fail. I think this will handle both cases:

```python
args = line[2:].rstrip().split(None, 3)
cmd = args.pop(0)
if cmd == '/usr/bin/env' and args:
    prog = args.pop(0)
    if prog == '-i' and args:
        progs = args.pop(0)
    cmd = '(env)/%s' % prog
```

dm0-: As far as I know, Linux doesn't support multiple arguments on the `#!` line.  I can still update it anyway, for portability (Container Illumos?).

dm0-: Updated.

marineam: Oh, right. forgot about that.

PR #620, opened 2016-12-29
set_lsb_release: update name and color

crawford:
/cc @philips @robszumski @meghanschofield

### DIFF
--- a/build_library/set_lsb_release
+++ b/build_library/set_lsb_release
@@ -25,9 +25,9 @@ ROOT_FS_DIR="$FLAGS_root"
 [ -n "$ROOT_FS_DIR" ] || die "--root is required."
 [ -d "$ROOT_FS_DIR" ] || die "Root FS does not exist? ($ROOT_FS_DIR)"
 
-OS_NAME="CoreOS"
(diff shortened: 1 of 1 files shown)


build_library/set_lsb_release
```diff
@@ -25,9 +25,9 @@ ROOT_FS_DIR="$FLAGS_root"
 [ -n "$ROOT_FS_DIR" ] || die "--root is required."
 [ -d "$ROOT_FS_DIR" ] || die "Root FS does not exist? ($ROOT_FS_DIR)"
 
-OS_NAME="CoreOS"
+OS_NAME="CoreOS Container Linux"
 OS_CODENAME="Ladybug"
-OS_ID="${OS_NAME,,}" # lower case for os-release
+OS_ID="containerlinux"
```
marineam: So... thoughts on what distro detection code we are going to break? e.g. old OEMs w/ the patched python that added support for using os-release? What about user code?

crawford: rrrrrrrrrrrrrrrrrrrrrrrrrr, you're right. coreos4lyfe

PR #630, opened 2017-01-12
coreos_sign_update: Use smartcards for signing

mjg59:
Sign updates using private keys on smartcards. This involves changing the
padding approach - rather than including the padding in the hash, ask the
card to generate the padding itself, since the card will refuse to sign
pre-padded material. Use + as a key separator rather than : as the PKCS#11
URI includes colons.

### DIFF
--- a/core_sign_update
+++ b/core_sign_update
@@ -43,10 +43,10 @@ delta_generator \
     -new_kernel "$FLAGS_kernel" \
     -out_file update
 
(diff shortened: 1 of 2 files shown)


mjg59:
Don't merge until ready to use smartcards.

euank:
Closing in favour of #710 which has a few additional changes.

PR #636, opened 2017-02-02
build_image: Extract and upload GRUB/shim EFI images for signing

bgilbert:
On arm64, extract only GRUB, since there is no shim.  On dev builds, extract neither.

Tested with:
- amd64 dev
- amd64 prod
- arm64 prod
- amd64 prod official

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -315,6 +315,8 @@ finish_image() {
   local image_contents="$4"
   local image_kernel="$5"
   local pcr_policy="$6"
+  local image_grub="$7"
+  local image_shim="$8"
 
   local install_grub=0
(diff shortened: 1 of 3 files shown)


build_library/build_image_util.sh
```diff
@@ -429,20 +431,24 @@ finish_image() {
     if [[ ${BOARD} == "arm64-usr" ]]; then
       target_list="arm64-efi"
     fi
+    local grub_args=()
+    if [[ ${FLAGS_enable_rootfs_verification} -eq ${FLAGS_TRUE} ]]; then
```
dm0-: Might as well use `${disable_read_write}` in preparation for #637.

bgilbert:
Rebased.

PR #637, opened 2017-02-02
Fix verity in our automated builds

dm0-:


crawford:
What is the error that this is fixing?

dm0-:
Verity is completely disabled in all of our automated builds, including the current alpha.

PR #653, opened 2017-03-02
*: don't hardcode path to bash

crawford:


bgilbert:
This only changes a small subset of the scripts that hardcode /bin, so not really `*`.  What problem is this solving?

euank:
No problems here, only solutions



@crawford shouldn't that path be fine within the sdk chroot though, where we can make such assumptions?

bgilbert:
I see.  @crawford, if you proceed with this change, please add an explanation to the commit message.  Otherwise this will be incomprehensible to anyone looking at the commit history.

crawford:
@euank these scripts are called outside of the SDK. The others are fine assuming the bash location.

PR #655, opened 2017-03-10
os-release: Add COREOS_BOARD variable

glevand:
os-release is requested in bug reports, and knowing which board
the problem occurred on is often helpful.


bgilbert:
[os-release(5)](https://www.freedesktop.org/software/systemd/man/os-release.html) asks that we namespace custom fields (in the last paragraph of the Options section).  How about `COREOS_BOARD`?

glevand:
Changed `BOARD` to `COREOS_BOARD`.  Also moved  `COREOS_BOARD` to the bottom of the file to be below the standard entries.

PR #656, opened 2017-03-15
build_library: upload image licenses

crawford:
We've always generated these license manifests (detailing which ebuilds
are covered by which license), but never published them. This adds these
manifests to the list of published files so that they are publicly
available.


marineam:
Oops, didn't realize I missed this :(

PR #647, opened 2017-02-27
Fixups for jenkins

glevand:
For developer builds we weren't setting COREOS_BUILD_ID and PORTAGE_BINHOST correctly.  Also, remove the old jenkins jobs that have been replaced by https://github.com/coreos/jenkins-os.

PR #659, opened 2017-03-29
common: Don't export COREOS_BUILD_ID

bgilbert:
https://github.com/coreos/scripts/pull/647 started exporting COREOS_BUILD_ID whenever it was found in version.txt, even if its value was blank. Because COREOS_BUILD_ID is in ENVIRONMENT_WHITELIST, this caused generated build IDs to be propagated into the SDK chroot environment and reused for every build in a `cork enter` session. Stop exporting COREOS_BUILD_ID when we set it ourselves.

See also #550.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -296,6 +296,10 @@ if [[ -f "${REPO_MANIFESTS_DIR}/version.txt" ]]; then
   else
     load_environment_var "${REPO_MANIFESTS_DIR}/version.txt" \
     COREOS_VERSION_ID COREOS_BUILD_ID COREOS_SDK_VERSION
+    # Don't promote COREOS_BUILD_ID into an environment variable when it
+    # didn't start as one, since we don't want it leaking into the SDK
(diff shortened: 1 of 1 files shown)


common.sh
```diff
@@ -296,6 +296,10 @@ if [[ -f "${REPO_MANIFESTS_DIR}/version.txt" ]]; then
   else
     load_environment_var "${REPO_MANIFESTS_DIR}/version.txt" \
     COREOS_VERSION_ID COREOS_BUILD_ID COREOS_SDK_VERSION
+    # Don't promote COREOS_BUILD_ID into an environment variable when it
+    # didn't start as one, since we don't want it leaking into the SDK
+    # chroot environment via ENVIRONMENT_WHITELIST.
+    typeset +x COREOS_BUILD_ID
```
dm0-: I think this should be `declare` instead of `typeset` to make bash happy, but there is no functional difference, so LGTM.

bgilbert: Yep, updated.

PR #660, opened 2017-03-30
Sign uploaded binary packages

dm0-:
This is a work in progress.  There is an issue with `emerge` commands failing unless `PORTAGE_SSH_OPTS=` is set in the environment.  The was this is set up, it only works with the SDK scripts since the `FETCHCOMMAND_GS` and `RESUMECOMMAND_GS` variables are not in the user's login environment.

### DIFF
--- a/build_library/release_util.sh
+++ b/build_library/release_util.sh
@@ -110,6 +110,7 @@ sign_and_upload_files() {
     local sigs=()
     if [[ -n "${FLAGS_sign}" ]]; then
         local file
+        local sigfile
         local sigdir=$(mktemp --directory)
         trap "rm -rf ${sigdir}" RETURN
         for file in "$@"; do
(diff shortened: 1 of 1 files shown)


dm0-:
It may be worth dropping the final commit (for downloading+verifying), and just taking the commits for signing packages to allow builds to start uploading signatures.

build_library/release_util.sh
```diff
@@ -110,15 +110,25 @@ sign_and_upload_files() {
...
             if [[ "${file}" =~ \.(asc|gpg|sig)$ ]]; then
                 continue
             fi
 
-            rm -f "${file}.sig"
-            gpg --batch --local-user "${FLAGS_sign}" \
-                --detach-sign "${file}" || die "gpg failed"
-            sigs+=( "${file}.sig" )
+            for sigfile in $(find "${file}" ! -type d); do
+                rm -f "${sigdir}/${sigfile}.sig"
```
dm0-: Updated

dm0-:
The download/verify part has been dropped for now.

PR #667, opened 2017-04-08
oem/ami: Drop in favor of plume

bgilbert:


bgilbert:
I've verified that everything in these scripts (except for running tests) is implemented in https://github.com/coreos/mantle/pull/540.

PR #680, opened 2017-05-04
Change default value of dhcp for vmware template to no

xcompass:
When OVA template is not being used, the default dhcp value yes will
trigger cloud-init to generate a 00-.network file, which will break
network connectivity Intermittently. Please see the details here:
https://github.com/coreos/bugs/issues/1802#issuecomment-297847614


bgilbert:
Tested with VMware Workstation 12.5 + `vapprun`; LGTM.

PR #685, opened 2017-05-19
Populate the torcx vendor store in OS images

dm0-:
Part of coreos/coreos-overlay#2567.

### DIFF
--- a/build_torcx_store
+++ b/build_torcx_store
@@ -1,0 +1,174 @@
+#!/bin/bash
+
+# Copyright (c) 2017 The CoreOS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+. "$(dirname "$0")/common.sh" || exit 1
+
+# Script must run inside the chroot
+restart_in_chroot_if_needed "$@"
+
+assert_not_root_user
+
+# Developer-visible flags.
+DEFINE_string board "${DEFAULT_BOARD}" \
+  "The board to build packages for."
+DEFINE_string output_root "${DEFAULT_BUILD_ROOT}/torcx" \
+  "Directory in which to place torcx stores (named by board/version)"
+
+# include upload options
+. "${BUILD_LIBRARY_DIR}/release_util.sh" || exit 1
+
+FLAGS_HELP="usage: $(basename $0) [flags] [images]
+
+This script builds a collection of torcx images to be installed into a torcx
+store.  By default, all supported images are built, but a list of images can be
+given as command arguments.  Note that their order matters, since the version
+specified last will get the default reference symlink.
+"
+show_help_if_requested "$@"
+
+# The following options are advanced options, only available to those willing
+# to read the source code. They are not shown in help output, since they are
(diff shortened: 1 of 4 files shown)


build_torcx_store
```diff
@@ -0,0 +1,176 @@
...
+
+# The following options are advanced options, only available to those willing
+# to read the source code. They are not shown in help output, since they are
+# not needed for the typical developer workflow.
+DEFINE_integer build_attempt 1 \
+  "The build attempt for this image build."
+DEFINE_string group developer \
+  "The update group."
+DEFINE_string version '' \
+  "Overrides version number in name to this version."
```
dm0-: I assume it's referring to the name of the directory it creates under `build`, which is the only reason I included that flag.  (I just copied it from `build_image` and left the description.)

build_torcx_store
```diff
@@ -0,0 +1,176 @@
...
+. "${BUILD_LIBRARY_DIR}/toolchain_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/build_image_util.sh" || exit 1
+
+# Print the first level of runtime dependencies for a torcx meta-package.
+function torcx_dependencies() (
+        local pkg=${1:?}
+        local ebuild inherit
+        ebuild=$(equery-${BOARD} w "${pkg}")
+        function inherit() { : ; }
```
dm0-: I can drop `inherit` from `local` if it's confusing, but it's in a subshell, so it shouldn't be global anyway.

build_torcx_store
```diff
@@ -0,0 +1,176 @@
...
+
+# Build and install a package configured as part of a torcx image.
+function torcx_build() (
+        local pkg=${2:?}
+        local tmproot=${1:?}
+
+        export LDFLAGS=-Wl,-rpath,/ORIGIN/../lib
+        export PKGDIR="${tmproot}/var/lib/portage/pkgs"
+
+        # Allow the meta-package to install bashrc to customize the builds.
```
dm0-: It is in a subshell as part of the function definition.

dm0-: Oh, and there isn't a reason for it at the moment with only Docker packages.  But we don't have the equivalent of real portage profiles here to support `/etc/portage/bashrc` since we're using the regular board profiles.  We can add a subprofile for this that inherits the board profile, but that will increase complexity again.  The idea behind a `bashrc` here is to support different `INSTALL_MASK`s etc. that must differ for packages between torcx and prod boards.

dm0-:
And an open question (maybe for @lucab or @euank):  Should we drop the Docker 17.05 torcx package from this for the current plan where we leave Docker 17.05 in `/usr`?

build_torcx_store
bgilbert: > It is in a subshell as part of the function definition.

That is an exceptionally subtle trick I hadn't seen before.  :+1:

build_torcx_store
bgilbert: Yeah, I'd drop from `local`.

PR #692, opened 2017-06-14
Use new oem packages for virtualbox and vagrant

AlexNPavel:
These commits set the oem package for virtualbox to "virtualbox" and vagrant to "vagrant-virtualbox". Depends on https://github.com/coreos/coreos-overlay/pull/2602 and https://github.com/coreos/coreos-overlay/pull/2603

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -55,6 +55,7 @@ VALID_OEM_PACKAGES=(
     rackspace-onmetal
     vagrant
     vagrant-key
+    virtualbox
     vmware
     niftycloud
     cloudsigma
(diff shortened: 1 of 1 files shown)


build_library/vm_image_util.sh
```diff
@@ -54,7 +54,9 @@ VALID_OEM_PACKAGES=(
     rackspace
     rackspace-onmetal
     vagrant
+    vagrant-virtualbox
```
bgilbert: sort order

AlexNPavel: Fixed

crawford:
Can you squash the commits so that there aren't any corrections in the history? Generally, I will leave these around and prefix them with "review: " if the review is huge and I don't want people to rereview lots of code. Once it's accepted, I'll squash the commits appropriately. Since this commit is small enough, I wouldn't bother with that.

AlexNPavel:
The last 2 commits were squashed

PR #704, opened 2017-06-27
build_packages: break gnupg dependency loop

dm0-:
Follow-up to coreos/portage-stable#560.


dm0-:
Or, as an alternative to this, we can just disable `smartcard` support in GPG again.

dm0-:
We'll use coreos/coreos-overlay#2630 to not add more packages to production images.

PR #710, opened 2017-07-06
coreos_sign_update: Use smartcards for signing (updated)

euank:
This replaces #630

See the additional commits for changes made.

This has been lightly tested by re-signing a previous release and getting an update payload with the same hash.

cc @crawford @dm0-

### DIFF
--- a/core_sign_update
+++ b/core_sign_update
@@ -43,10 +43,10 @@ delta_generator \
     -new_kernel "$FLAGS_kernel" \
     -out_file update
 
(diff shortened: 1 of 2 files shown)


euank:
Adding a flag for the separator seemed cleaner to me. Per the reasoning in the commit message I think backwards compatibility is worth keeping for now.

euank:
The commit message might not have been quite clear enough on that. The intent is that if something's wrong with the new setup for at least the next few uses, we can trivially go back to the existing one without having to fiddle with script changes (even if they're minor).

PR #715, opened 2017-07-26
Support dropping Python 3 in alpha

dm0-:
This is to run test builds for Jenkins SDKs downgrading to releases without Python 3.


dm0-:
A test build is still running, but it corrected the issues so far.

PR #721, opened 2017-07-28
build_library: Update qemu script to run on arm64 machines

glevand:
I used a condensed indentation in the `case` statements to avoid the `set` commands from wrapping so much.  I can change it to match what was there before if you like.


glevand:
Changed case to `"${VM_BOARD}+$(uname -m)"`.
Added new patch `	build_library: Add missing die function`.

PR #722, opened 2017-08-08
oem/oracle: Add some scripts to manage images/instances

bgilbert:


oem/oracle/get-console-output
```diff
@@ -0,0 +1,35 @@
...
+if [[ -z "${instance_id}" ]]; then
+    echo "Instance ID is required." >&2
+    exit 2
+fi
+
+id=$(bmcs compute console-history capture --instance-id "${instance_id}" | jq -r .data.id)
+trap 'bmcs compute console-history delete --instance-console-history-id "${id}" --force' EXIT
+while true; do
+    state=$(bmcs compute console-history get --instance-console-history-id "${id}" | jq -r '.data["lifecycle-state"]')
+    if [[ "${state}" = SUCCEEDED ]]; then
```
arithx: `SUCCEEDED` isn't wrapped in quotes

oem/oracle/get-console-output
```diff
@@ -0,0 +1,35 @@
...
+    esac
+done
+
+if [[ -z "${instance_id}" ]]; then
+    echo "Instance ID is required." >&2
+    exit 2
+fi
+
+id=$(bmcs compute console-history capture --instance-id "${instance_id}" | jq -r .data.id)
+trap 'bmcs compute console-history delete --instance-console-history-id "${id}" --force' EXIT
```
arithx: Why are you deleting the console output on exit?

oem/oracle/upload-image
```diff
@@ -0,0 +1,75 @@
...
+    echo "Display name is required." >&2
+    exit 2
+fi
+
+if [[ -z "${path}" ]]; then
+    echo "Image file is required." >&2
+    exit 2
+fi
+
+object=$(mktemp -u XXXXXXXXXXXXXXXX)
```
arithx: no trap to delete this

bgilbert: `-u` doesn't create a file; we're just generating a random string here.

oem/oracle/get-console-output
bgilbert: There's no point leaving console history objects sitting around in the compartment once we've fetched them.

oem/oracle/get-console-output
bgilbert: Constant strings don't need to be.  Sometimes they are as a stylistic thing, but I don't think we're rigid about it.

PR #729, opened 2017-08-15
build_library: Add Vagrant box files

bgilbert:
No longer have the OEM install them in the OEM partition only for `vm_image_util.sh` to move them back out. For now this only applies to the `vagrant_virtualbox` image, but the goal is to migrate all the Vagrant
images and drop the FS hook.

Box files copied from coreos/coreos-overlay@bd0cd7685ebf71a133c1b70f5ebd5a0b3393ac70 `coreos-base/oem-vagrant-virtualbox/files/box`.

cc @AlexNPavel

### DIFF
--- a/build_library/vagrant/Vagrantfile
+++ b/build_library/vagrant/Vagrantfile
@@ -1,0 +1,31 @@
+# -*- mode: ruby -*-
+# # vi: set ft=ruby :
+
+if Vagrant::VERSION < "1.6.0"
+  raise "Need at least vagrant version 1.6.0, please update"
+end
+
+require_relative 'change_host_name.rb'
+require_relative 'configure_networks.rb'
+require_relative 'base_mac.rb'
+
+Vagrant.configure("2") do |config|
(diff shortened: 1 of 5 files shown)


AlexNPavel:
LGTM. I guess the only other change required would be to make this to use the oem-virtualbox package right? At the moment, we still need the `vagrant-virtualbox` oem-id, but we're in the process of migrating all usage of that to the `virtualbox` id instead, so we can do that once all those changes are ready (the oem-id and oem-release file are the only differences between the 2 oem packages from what I can tell).

bgilbert:
Thank you for submitting this PR. Unfortunately, we don't think we'll end up addressing it in Container Linux.

We're now working on [Fedora CoreOS](https://coreos.fedoraproject.org/), the successor to Container Linux, and we expect most major development to occur there instead. Meanwhile, Container Linux will be fully maintained into 2020 but won't see many new features. We appreciate your taking the time to submit this PR and we're sorry that we won't be able to address it.

PR #730, opened 2017-08-15
jenkins: Upload a list of default torcx images

dm0-:
Completely untested

### DIFF
--- a/jenkins/images.sh
+++ b/jenkins/images.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -ex
 
 # Clear out old images.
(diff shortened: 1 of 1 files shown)


dm0-:
I used this over the nightly build's run, and it looks okay.

euank:
Not including the patchlevel version isn't great.. and after further thought, I think we need to have a little more discussion around what we're doing.

Punting the decision forwards to the next alpha.

dm0-:
Sounds like a more involved solution is being designed, so closing this.

PR #733, opened 2017-08-30
scripts: add more options in qemu script

cristiklein:
scripts: add more options in qemu script

I frequently need to change the number of processors, memory and VM name. This change allows the user to specify these options when booting CoreOS from the qemu script.

### DIFF
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -53,6 +53,15 @@ check_conflict() {
 
 while [ $# -ge 1 ]; do
     case "$1" in
+        -vm-name)
+            VM_NAME="$2"
+            shift 2 ;; 
+        -ncpus)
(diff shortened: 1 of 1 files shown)


euank:
Trailing arguments are supported and added to the qemu invocation.

Using something like:

```sh
$ ./coreos_production_qemu.sh -name asdf -smp cpus=10
```

will change the name, number of processors, etc.

I'm not sure we need this mechanism as well since it doesn't hide a significant amount of abstraction (as, for example, the ignition flag does).

cristiklein:
I see your point. There is, however, a minor downside to the solution you proposed, in that the command-line of qemu (e.g., `ps -AF`) looks confusing.

Would you be more open to an `-image` option to allow pointing to a COW-image of CoreOS instead of the one shipped by default?

PR #735, opened 2017-09-06
offline_signing: Allow download/upload for multiple releases with a single invocation

bgilbert:


bgilbert:
I run `download` outside the SDK.  On Fedora, 2.2.0 is in rawhide but only as `gpg2`.

PR #737, opened 2017-09-06
rework torcx uploading to include a manifest

euank:
- [x] `./build_packages` produces manifest and cas pkgs
- [x] `./build_packages --upload` does the right thing for developers
- [x] `./build_image` on dev machines includes torcx packages based on a manifest correctly
- [x] `./jenkins/images.sh` correctly populates torcx store and can `build_image`
- [x] `kola` accepts a 'torcx manifest' path to test a manifest prior to signing / upload (https://github.com/coreos/mantle/pull/701)
- [ ] signing scripts for download/signing/uploading the torcx manifest

PR #742, opened 2017-09-09
Backport build_torcx_store manifest changes

euank:
This backports https://github.com/coreos/scripts/pull/737

We'll want to pull in https://github.com/euank/scripts/commit/e858e727b05668da98ebb74878c298097cf2a03b once this branch goes to beta as well.


euank:
I manually gave `./build_torcx_store --upload` a run and it seemed to work.

Since we aren't using `casync` any more, there isn't any need for overlay / portage changes.

dm0-:
Shouldn't it still build 17.06 when it goes to beta?  We don't want it as an option?

euank:
@dm0- the tectonic-torcx thing only really needs 1.12, but if we want to toss 17.06 in EXTRA_IMAGES I wouldn't say no. It just won't immediately be used other than in testing.

PR #746, opened 2017-09-14
offline_signing: Mechanize more roller interactions

bgilbert:
Also parallelize download of artifacts for signing, mostly to stop gsutil from complaining that we didn't.

### DIFF
--- a/offline_signing/transfer.sh
+++ b/offline_signing/transfer.sh
@@ -21,7 +21,7 @@ download() {
     mkdir -p "${dir}"
     pushd "${dir}" >/dev/null
 
-    gsutil cp \
+    gsutil -m cp \
         "${gs}/coreos_production_image.vmlinuz" \
         "${gs}/coreos_production_image.vmlinuz.sig" \
         "${gs}/coreos_production_update.bin.bz2" \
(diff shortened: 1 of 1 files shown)


offline_signing/transfer.sh
```diff
@@ -49,6 +49,19 @@ upload() {
         --board="${board}" \
         --version="${version}" \
         --payload="${payload}"
+
+    # Update version in a canary channel if one is defined.
+    local channel_var="ROLLER_CANARY_CHANNEL_$(tr a-z A-Z <<<${channel})"
```
dm0-: Minor nit, but maybe helps readability?  This would allow using it as a regular variable without the `!` indirection:

    local -n canary_channel="ROLLER_CANARY_CHANNEL_${channel^^}"

offline_signing/transfer.sh
```diff
@@ -6,6 +6,11 @@ declare -A APPID
 APPID[amd64-usr]=e96281a6-d1af-4bde-9a0a-97b76e56dc57
 APPID[arm64-usr]=103867da-e3a2-4c92-b0b3-7fbd7f7d8b71
 
+declare -A RELEASE_CHANNEL
```
dm0-: If we're defining these channel names, would it make sense to just map the canary channels as well?  Is there a reason we'd need to change them through those environment variables?

bgilbert: That was done solely to avoid publishing the channel names.  The variable assignments can go into `~/.config/roller.conf` and the internal release docs will specify their values.

dm0-: Okay, I forgot those aren't published.  Makes sense then.

offline_signing/transfer.sh
bgilbert: Two tricks I didn't know about.  Will fix.

PR #747, opened 2017-09-15
build_library/torcx_manifest: omit newlines

euank:
Omit the newline for each function that returns only a single value.

This does not change each function that returns multiple values as those
are expected, by the caller, to be newline delimited values.

Before this change, the consumer of 'local_store_path' specifically
ended up with bad behavior of including a newline in the on-disk
filename.

Fortunately, the symlink included that newline as well, so it didn't
*break* things, but it did mean the filepath differed from the one
declared in the manifest slightly.

Pending a full `./build_package && ./bulid_image` cycle to verify this didn't break anything obvious.

### DIFF
--- a/build_library/torcx_manifest.sh
+++ b/build_library/torcx_manifest.sh
@@ -96,7 +96,7 @@ function torcx_manifest::local_store_path() {
   local file="${1}"
   local name="${2}"
   local version="${3}"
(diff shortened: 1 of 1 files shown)


euank:
I don't think, after some manual fiddling, this was actually the issue.

PR #749, opened 2017-09-15
build_library: Filter out null paths in torcx manifests

dm0-:
This avoids odd "null" lines appearing in torcx image names.

### DIFF
--- a/build_library/torcx_manifest.sh
+++ b/build_library/torcx_manifest.sh
@@ -96,7 +96,7 @@ function torcx_manifest::local_store_path() {
   local file="${1}"
   local name="${2}"
   local version="${3}"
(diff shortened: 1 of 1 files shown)


build_library/torcx_manifest.sh
```diff
@@ -96,7 +96,7 @@ function torcx_manifest::local_store_path() {
   local file="${1}"
   local name="${2}"
   local version="${3}"
-  jq -r ".value.packages[] | select(.name == \"${name}\") | .versions[] | select(.version == \"${version}\") | .locations[].path" < "${file}"
+  jq -r ".value.packages[] | select(.name == \"${name}\") | .versions[] | select(.version == \"${version}\") | .locations[] | select(.path) | .path" < "${file}"
```
euank: total nit, but `select(.path).path` is equivalent and matches what's done elsewhere.

PR #752, opened 2017-09-19
update_chroot: Avoid a portage crash while upgrading binutils

dm0-:
This omits the toolchain packages' version-pinning flag for the binutils package while it is being upgraded.  It also removes older versions installed in parallel that cause unwanted rebuilds.

When stable has the upgraded version, this can be reverted.

Part of coreos/portage-stable#586.

### DIFF
--- a/build_library/toolchain_util.sh
+++ b/build_library/toolchain_util.sh
@@ -194,7 +194,9 @@ get_cross_pkgs() {
 # Get portage arguments restricting toolchains to binary packages only.
 get_binonly_args() {
     local pkgs=( "${TOOLCHAIN_PKGS[@]}" $(get_cross_pkgs "$@") )
-    echo "${pkgs[@]/#/--useoldpkg-atoms=}" "${pkgs[@]/#/--rebuild-exclude=}"
+    # XXX: Revert this binutils change after stable > 1535.
+    local tmppkgs=( ${pkgs[*]##*/binutils} )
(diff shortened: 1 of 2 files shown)


build_library/toolchain_util.sh
```diff
@@ -194,7 +194,9 @@ get_cross_pkgs() {
 # Get portage arguments restricting toolchains to binary packages only.
 get_binonly_args() {
     local pkgs=( "${TOOLCHAIN_PKGS[@]}" $(get_cross_pkgs "$@") )
-    echo "${pkgs[@]/#/--useoldpkg-atoms=}" "${pkgs[@]/#/--rebuild-exclude=}"
+    # XXX: Revert this binutils change after stable > 1535.
+    local tmppkgs=( "${TOOLCHAIN_PKGS[@]#sys-devel/binutils*}" $(get_cross_pkgs "$@") )
```
euank: I think just deleting or commenting out the line in the TOOLCHAIN_PKGS variable would be more readable. Would that not work for some reason?

dm0-: This doesn't work at all yet for amd64 in automated builds.  I am still working on it.

euank:
Including this, I end up running into:

```

dm0-:
This worked in Jenkins builds, I haven't tried it locally for a while.

euank:
Turns out it was my amd64-usr copy of binutils it was complaining about. After manually depcleaning that, this seems to work fine.

PR #763, opened 2017-10-09
update_ebuilds: switch to 'gentoo-mirror'

euank:
This mirror repository contains additional metadata, including all GLSAs
and so on.

This also specifies just the 'stable' branch for fetching from since
there are a *lot* of branches in that repo.

cc @dm0-, looks like I picked the wrong repo last time


euank:
It looks like this still drops the ChangeLog file, which is desirable, but doesn't really matter either way.

PR #768, opened 2017-10-24
build_torcx_store: Upgrade to Docker 17.10 on disk

dm0-:
Part of coreos/coreos-overlay#2851


dm0-:
Updated to build both 17.10 and 17.09 for automated testing.

PR #769, opened 2017-10-25
WIP: check_out_of_date: add script

ajeddeloh:
Add a script to check where portage-stable or an image package list is
out of date with regards to gentoo upstream.

Add a script to take tab delimited tables and create html tables.


ajeddeloh:
Updated to actually clone in upstream if it doesn't exist and generally cleaned it up.

ajeddeloh:
Updated to use a little less sed, take into account stage1 and stage2 catalyst packages.

ajeddeloh:
Depends on https://github.com/coreos/coreos-overlay/pull/2867

find_overlay_dups
```diff
@@ -0,0 +1,35 @@
...
+function get_tree_packages() {
+	# gets a list of all packages in a tree
+	find "$1" -type f -name "*.ebuild" -printf "%P\n" | xargs dirname | uniq
+}
+
+portage_stable_packages=$(get_tree_packages ${FLAGS_portage_stable_path})
+overlay_packages=$(get_tree_packages ${FLAGS_overlay_path})
+
+all_packages="$portage_stable_packages $overlay_packages"
+dups=$(sort <<< "$all_packages" | uniq -D | uniq)
```
dm0-: If you want to avoid storing long lists and writing a tmpfile, I think this does the same thing:

    find -L "${FLAGS_portage_stable_path}" "${FLAGS_overlay_path}" -maxdepth 3 -type f -name '*.ebuild' -printf '%P\n' | sed s,/[^/]*$,, | uniq | sort | uniq -d

check_out_of_date
```diff
@@ -0,0 +1,271 @@
...
+function get_versions() {
+	local package="$1"
+	local local_vers=$(get_versions_from_pkg_dir "${portage_stable_path}/${package}")
+	packages[$package]="${local_vers}"
+	get_upstream_versions "${package}"
+}
+
+function get_upstream_versions() {
+	local package="$1"
+	local package_dir="${gentoo_path}/${package}"
```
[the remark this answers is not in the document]
ajeddeloh: I had it clone the directory so if you're running it multiple times you only need to copy everything once (i.e. you don't do a clone/pull every time). We could use the `ls-files` and write it's output to a file though, which would be a little lighter.

check_out_of_date
```diff
@@ -0,0 +1,271 @@
...
+		get_upstream_versions ${package}
+	done
+}
+
+function get_package_date() {
+	local package="$1"
+	pushd ${portage_stable_path} > /dev/null
+	git_date=$(git log -1 --pretty=format:%ad --date="${FLAGS_date_fmt}" ${package} | cat)
+	popd > /dev/null
+	echo "${git_date}"
```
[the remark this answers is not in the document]
ajeddeloh: yup, I didn't know about `git -C`

check_out_of_date
```diff
@@ -0,0 +1,271 @@
...
+DEFINE_boolean colors ${FLAGS_TRUE} \
+    "Output colorfully"
+DEFINE_boolean use_image ${FLAGS_FALSE} \
+    "Check versions in the last built image instead of portage-stable"
+DEFINE_string package_list_path "${DEFAULT_BUILD_ROOT}/images/amd64-usr/latest/coreos_production_image_packages.txt" \
+    "Path the image package list if use_image is true"
+DEFINE_string date_fmt "relative" \
+    "what date format to use with git log"
+DEFINE_boolean use_html ${FLAGS_FALSE} \
+    "Whether to output html"
```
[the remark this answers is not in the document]
ajeddeloh: The idea was to have jenkins generate something human readable. I'm not married to the implementation for sure.

find_overlay_dups
ajeddeloh: Does find always do a DFS or is that an implemenation detail. I don't see anything in `man find` defining a traversal order.

ajeddeloh:
Rewrote the check_out_of_date in python. I got sick of doing set operations in bash and hey, now we get json output as well.

find_overlay_dups
dm0-: The man page says the `depth` option uses depth-first ordering, although I don't think that affects this case.  All the ebuilds would be at the same depth.

ajeddeloh: :+1: Although if a package is last in `portage-stable` and first in `coreos-overlay`, it would be skipped.

arithx:
Might be wortwhile to validate via `flake8` instead of `pylint` if we're not worried about a few things. Can disable the line length check via `flake8 --ignore E501`

ajeddeloh:
I'm 100% down for that. Updated to pass `flake8 --ignore E501` and it now has a pylint score of 5.9. We should look into enabling some CI for this, maybe require flake8 and just show pylint?

check_out_of_date.py
```diff
@@ -0,0 +1,186 @@
...
+                      "ours": ["Ours"],
+                      "upstream": ["Upstream"],
+                      "tag": "Tag",
+                      "sdk": ["sdk"],
+                      "arm64-usr": ["arm64-usr"],
+                      "amd64-usr": ["amd64-usr"],
+                      "bootstrap": ["bootstrap"],
+                      "modified": "Modified"})
+    for entry in report:
+        print(line_head + joiner.join([entry["name"],
```
arithx: Best practices would still have you use a `get` function even though this will always be defined in the context of this script. If this script was imported and used by something else it would error out with a `KeyError` if any of these fields were not defined.

ajeddeloh:
Updated to fix html, use `.get()` when printing tables.

ajeddeloh:
Review bump. Also fixed a bug where package names were not stripped correctly.

PR #770, opened 2017-11-07
build_packages: fix dependency loops

ajeddeloh:
Add more to the flags files to prevent systemd loops. Also support comma
syntax for arguments beyond the second argument.

### DIFF
--- a/build_packages
+++ b/build_packages
@@ -212,10 +212,11 @@ break_dep_loop() {
   # Disable any other problematic flags
   extra_args=""
   while [[ $# -gt 0 ]]; do
-      sudo_append "${flag_file}" <<<"$1 -$2"
+      sudo_append "${flag_file}" <<<"$1 -${2//,/ -}"
       extra_args+=" --buildpkg-exclude=$1 --useoldpkg-atoms=$1"
       shift 2
   done
+
   # rebuild-if-unbuilt is disabled to prevent portage from needlessly
   # rebuilding zlib for some unknown reason, in turn triggering more rebuilds.
   sudo -E "${EMERGE_CMD[@]}" "${EMERGE_FLAGS[@]}" \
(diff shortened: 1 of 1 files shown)


ajeddeloh:
Confirmed there are no more loops to fix.

build_packages
```diff
@@ -227,7 +228,10 @@ break_dep_loop() {
 
 if [[ "${FLAGS_usepkgonly}" -eq "${FLAGS_FALSE}" ]]; then
   # util-linux[udev] -> virtual->udev -> systemd -> util-linux
-  break_dep_loop sys-apps/util-linux udev,systemd sys-apps/systemd cryptsetup
+  break_dep_loop sys-apps/util-linux udev,systemd \
+    sys-apps/systemd udev,cryptsetup \
```
dm0-: There is no `udev` USE flag for systemd.

ajeddeloh:
@dm0 it was breaking new sdk's created via `cork create` and `cork enter`, run locally. I'll play around and see if I can get a more minimal set of flags here, hopefully only on `util-linux`.

dm0-:
What are the exact commands you run to reproduce it?  I've been wiping out Jenkins workspaces lately, so just about every nightly build starts with a fresh `cork` SDK for every job.

ajeddeloh:
```
mkdir coreos-test
cd coreos-test
cork create
cork enter
./setup_board --board=amd64-usr --default
./build_packages
```
I'll try again and see if the latest alpha being released fixed it.

dm0-:
I ran those commands just now and did not hit any errors.

ajeddeloh:
I also tried just now with (another) fresh sdk and it worked fine. But I definitely hit this and @euank repro'd. I guess I'll close this and reopen if it happens again?

PR #771, opened 2017-11-08
tag_release: by default, prompt before pushing

bgilbert:


tag_release
```diff
@@ -4,6 +4,22 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+git_confirm() {
+    local response
+    if [[ ${FLAGS_force} -ne ${FLAGS_TRUE} ]]; then
+        echo "Executing: git $@"
```
ajeddeloh: Is there a reason to use echo instead of info?

tag_release
```diff
@@ -4,6 +4,22 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+git_confirm() {
```
ajeddeloh: This seems useful to have in `common` as `exec_confirm` for future use. But since it appears this is the only user as of now, I'm also fine with it just living here until we need it elsewhere.

bgilbert: It assumes there is a `${FLAGS_force}` and that it's okay to `exit 1`. It might be useful to generalize but that doesn't seem completely obvious.

tag_release
bgilbert: It's not really a log message (e.g. it isn't shown if `--force` is given); it's part of the prompt.

tag_release
ajeddeloh: I mean, yeah it would need to change a little, but like I said, I'm not too concerned about it regardless.

PR #775, opened 2017-11-13
disk_util: use FAT32 on ESP

bgilbert:
`mkfs.vfat` was defaulting to FAT16 based on the size of the partition. The UEFI spec (2.7 errata A, section 13.3) implies that only FAT32 is necessarily supported on the ESP, and we've received a report of hardware that doesn't recognize FAT16.

Addresses https://github.com/coreos/bugs/issues/2246.


ajeddeloh:
What's the implication of this with systems updating. I'm assuming we'll end up with people provisioning new machines having fat32 and old machines having fat16?

bgilbert:
Sadly, yes.

PR #776, opened 2017-11-22
build_torcx_store: Update the default Docker to 17.11

dm0-:
Part of coreos/coreos-overlay#2914


dm0-:
I've added commits to package `/usr/share` for the containerd config file.  That also brings in Docker's vimrc files, but nothing else.  It didn't seem like enough to warrant adding masking.

This also sets a variable on service files with the subdirectory under `$TORCX_UNPACKDIR` where the image's files can be found, since that location isn't known anywhere else.

PR #777, opened 2017-12-02
build_torcx_store: Make the LTS Docker CE version the default

dm0-:
Prepare alpha for beta promotion.

### DIFF
--- a/build_torcx_store
+++ b/build_torcx_store
@@ -219,14 +219,13 @@ function torcx_package() {
 # swapping default package versions for different OS releases by reordering.
 DEFAULT_IMAGES=(
         =app-torcx/docker-1.12
-        =app-torcx/docker-17.10
+        =app-torcx/docker-17.09
 )
 
 # This list contains extra images which will be uploaded and included in the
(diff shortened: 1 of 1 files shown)


dm0-:
They just posted a 17.09.1 RC, so I'll wait until the last minute to merge in case that gets released before beta is built.

build_torcx_store
```diff
@@ -219,14 +219,14 @@ function torcx_package() {
...
-        =app-torcx/docker-17.10
+        =app-torcx/docker-17.09
 )
 
 # This list contains extra images which will be uploaded and included in the
 # generated manifest, but won't be included in the vendor store.
 EXTRA_IMAGES=(
 	=app-torcx/docker-17.03
-	=app-torcx/docker-17.09
+	=app-torcx/docker-17.10
```
bgilbert: Should we drop CE Edge entirely for beta?  It doesn't seem likely to be used, especially from the Tectonic-specific bucket.

dm0-: Okay, dropped.

PR #778, opened 2017-12-18
*: make prod images default

ajeddeloh:
Make prod images default instead of dev images. Additionally, clean up some of the flag logic around the --dev_image flag in image_to_vm.sh

This should probably wait until after 1632.0.0 because it will almost certainly break some jenkins jobs.

Edit: waiting until after 1646.0.0

@bgilbert


ajeddeloh:
Merging now that releases are tagged

PR #779, opened 2017-12-18
Revert "disk_util: use FAT32 on ESP"

ajeddeloh:
This reverts commit 7f058d61a10384ca81c88245c4ff80c08f89ad6b.

Reverting because of bug 2284 [1] where grub will sometimes fail due to
memory corruption. This is _not_ the cause of the bug, and the bug can
even be reproduced with this reversion, but it seems to occur less when
not using fat32.

[1] https://github.com/coreos/bugs/issues/2284


bgilbert:
Typo fix for clarity: `fail due to`.

ajeddeloh:
fixed s/do/due

PR #794, opened 2018-03-02
grub_install: Add LOOP_NO_UDEV for builds in a container

cgwalters:
A lot of the SDK instructions seem to assume you're in a "classic" Linux login
session. I use Fedora Atomic Workstation and do all of my development in "dev
containers" (currently docker, in the process of switching to podman).

In my setup the `/dev` setup is separate and won't pick up udev changes.
(Ideally I'd filter out a lot of host devices, that's another issue)

Anyone in a similar situation (which I assume would also include trying
to do the SDK build inside a Docker container on CoreOS) can do:
`env LOOP_NO_UDEV=1 ./build_images.sh`
to have it manually set up the partition mounts.

Loopback mounts with containers in general are ugly since they're not
namespaced.

A whole better solution to this IMO is to use something like
http://libguestfs.org/ which basically spawns a VM, although it doesn't support
grub2. So we'd really have to do instead something like what Fedora does with
using Anaconda. Or the "helper VM" could probably just be an existing CoreOS
qcow2.

### DIFF
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -52,6 +52,12 @@ CORE_NAME=
 # fixed up the board root's grub will always be used.
 BOARD_GRUB=0
 
+# This may be useful if you're running the build process inside a container
+# (e.g. docker/podman/etc) distinct from the host; in that case you'll have
+# a distinct /dev normally and won't pick up udev probed devices.  But if you
+# do the probe from the host side, it should show up in the container.
+LOOP_NO_UDEV=${LOOP_NO_UDEV:-0}
+
 case "${FLAGS_target}" in
     i386-pc)
         CORE_MODULES+=( biosdisk serial )
@@ -110,6 +116,21 @@ info "Installing GRUB ${FLAGS_target} in ${FLAGS_disk_image##*/}"
 LOOP_DEV=$(sudo losetup --find --show --partscan "${FLAGS_disk_image}")
 ESP_DIR=$(mktemp --directory)
 
+# Hack available for doing the build in a container; see the LOOP_NO_UDEV
+# comments above.
+if [[ ${LOOP_NO_UDEV} -eq 1 ]]; then
+    echo "LOOP_NO_UDEV enabled, creating device nodes directly"
+    LOOPNUM=$(echo ${LOOP_DEV} | sed -e 's,/dev/loop,,')
+    for d in /sys/block/loop${LOOPNUM}/loop${LOOPNUM}p*; do
+        p=$(cat ${d}/partition)
+        dev=$(cat ${d}/dev)
+        min=${dev##*:}
+        maj=${dev%:*}
+        devpath=/dev/loop${LOOPNUM}p${p}
+        sudo rm -f ${devpath}
+        sudo mknod -m 660 ${devpath} b ${maj} ${min}
+    done
+fi
 # work around slow/buggy udev, make sure the node is there before mounting
 if [[ ! -b "${LOOP_DEV}p1" ]]; then
     # sleep a little just in case udev is ok but just not finished yet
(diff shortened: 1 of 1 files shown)


euank:
As you point out, the SDK instructions do assume use of a normal login shell on a reasonably modern Linux.

This specific change doesn't seem very intrusive, but I suspect there may be other places in the sdk+tooling that are also broken (e.g. running qemu tests with `kola` produces network namespaces and may mess up).

The bigger issue though is the question of whether we want to support the use-case of running the SDK in a container. Until now, the answer has been implicitly "no" because afaik no one asked or tried.

I think answering that question makes it easier to figure out whether it makes sense for us to take this or for you to carry it.

I personally think it makes sense to begin to support it. This is in no small part because there's been discussion in the past about moving our non-release build/testing workloads onto Kubernetes, which would certainly be easier if the sdk runs well in a container. 

cc @dm0-, do you have an opinion either on this patch or on the broader question of whether we should start supporting use of the sdk in a container?

> Anyone in a similar situation (which I assume would also include trying
to do the SDK build inside a Docker container on CoreOS)

As a bit of an aside, `cork create` and `cork enter` work outside of a container on CL just fine, which is how we run our automated builds.

build_library/grub_install.sh
```diff
@@ -110,6 +116,21 @@ info "Installing GRUB ${FLAGS_target} in ${FLAGS_disk_image##*/}"
 LOOP_DEV=$(sudo losetup --find --show --partscan "${FLAGS_disk_image}")
 ESP_DIR=$(mktemp --directory)
 
+# Hack available for doing the build in a container; see the LOOP_NO_UDEV
+# comments above.
+if [[ ${LOOP_NO_UDEV} -eq 1 ]]; then
+    echo "LOOP_NO_UDEV enabled, creating device nodes directly"
+    LOOPNUM=$(echo ${LOOP_DEV} | sed -e 's,/dev/loop,,')
```
dm0-: These scripts are already full of bashisms, so we might as well cut out unnecessary processes: `LOOPNUM=${LOOP_DEV#/dev/loop}`

build_library/grub_install.sh
```diff
@@ -110,6 +116,21 @@ info "Installing GRUB ${FLAGS_target} in ${FLAGS_disk_image##*/}"
...
+    for d in /sys/block/loop${LOOPNUM}/loop${LOOPNUM}p*; do
+        p=$(cat ${d}/partition)
```
dm0-: `p=$(< ${d}/partition)`

build_library/grub_install.sh
```diff
@@ -110,6 +116,21 @@ info "Installing GRUB ${FLAGS_target} in ${FLAGS_disk_image##*/}"
...
+        dev=$(cat ${d}/dev)
```
dm0-: `dev=$(< ${d}/dev)`

build_library/grub_install.sh
```diff
@@ -110,6 +116,21 @@ info "Installing GRUB ${FLAGS_target} in ${FLAGS_disk_image##*/}"
 LOOP_DEV=$(sudo losetup --find --show --partscan "${FLAGS_disk_image}")
 ESP_DIR=$(mktemp --directory)
 
+# Hack available for doing the build in a container; see the LOOP_NO_UDEV
+# comments above.
+if [[ ${LOOP_NO_UDEV} -eq 1 ]]; then
+    echo "LOOP_NO_UDEV enabled, creating device nodes directly"
+    LOOPNUM=$(echo ${LOOP_DEV} | sed -e 's,/dev/loop,,')
+    for d in /sys/block/loop${LOOPNUM}/loop${LOOPNUM}p*; do
```
dm0-: For additional cleanliness, maybe have a `shopt -s nullglob` for this loop?  That way if something crazy happens that breaks this pattern, the rest of the script still functions normally when it checks for a nonexistent device node.  (This "library" script is only executed standalone rather than sourced, so a `shopt` shouldn't affect anything else.)

dm0-:
@euank I am in favor of supporting SDKs in a container--specifically shipping the SDK `chroot` directory as a container more than running `cork` from inside some other container.  We'd need to figure out how to handle the user mapping (or maybe just give up on it and always using `core`) and have optional volume mounts for passing in GCS creds, Git configuration, keyrings, agent sockets, and whatever else `cork` does.  Having a signed non-persistent SDK would be nice for our CI where we only want to run it once and upload the produced artifacts, and I'm sure there are some users who would prefer to be able to use their existing container setup rather than run our `cork` binary that has no other use on their systems.

PR #799, opened 2018-03-08
Fix a bug when setting VM_CDROM in qemu template

dongsupark:
Fix a bug in the qemu template, which is used for generating `coreos_production_qemu.sh` or similar. Setting `$VM_CDROM` in the qemu script does not work as expected when installing Container Linux from the given bootable CDROM image. That's probably because qemu-system-x86_64 expects another boot option `-boot order=d` to be able to boot from the given CDROM drive. Let's specify a `-boot` as well as `-drive` option for the given CDROM drive.

### DIFF
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -198,7 +198,8 @@ if [ -n "${VM_UUID}" ]; then
 fi
 
 if [ -n "${VM_CDROM}" ]; then
-    set -- -cdrom "${SCRIPT_DIR}/${VM_CDROM}" "$@"
+    set -- -boot order=d \
+	-drive file="${SCRIPT_DIR}/${VM_CDROM}",media=cdrom,format=raw "$@"
 fi
 
 if [ -n "${VM_PFLASH_RO}" ] && [ -n "${VM_PFLASH_RW}" ]; then


bgilbert:
The first commit looks okay.  I don't think the second one is a good idea.  If the qemu template needs additional options, it'd be better to add them as real options, rather than undocumented overrides of internal variables.

dongsupark:
@bgilbert I see.
I've just dropped the second commit. Anyway that's not a part of bugfix. I'll try to support it as real options in the future.
Thanks.

build_library/qemu_template.sh
```diff
@@ -198,7 +198,8 @@ if [ -n "${VM_UUID}" ]; then
 fi
 
 if [ -n "${VM_CDROM}" ]; then
-    set -- -cdrom "${SCRIPT_DIR}/${VM_CDROM}" "$@"
+    set -- -boot once=d \
```
bgilbert: Looking at this more closely, this should be `order=d` rather than `once=d`.  Otherwise rebooting the machine will fail if a hard disk is specified on the command line.

dongsupark: @bgilbert Done. I changed it to `order=d`.

build_library/qemu_template.sh
```diff
@@ -198,7 +198,8 @@ if [ -n "${VM_UUID}" ]; then
 fi
 
 if [ -n "${VM_CDROM}" ]; then
-    set -- -cdrom "${SCRIPT_DIR}/${VM_CDROM}" "$@"
+    set -- -boot order=d \
+	-drive file=${SCRIPT_DIR}/${VM_CDROM},media=cdrom,format=raw "$@"
```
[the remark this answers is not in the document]
dongsupark: @bgilbert Right. I've just added the quotes.

PR #800, opened 2018-03-22
[WIP] release_util: support a smartcard pin for signing

euank:
This adds support to 'sign_and_upload_files' for a smartcard pin and
furthermore plumbs it through all the jenkins bits.

This is untested as of yet.


euank:
The jenkins bit ends up being a bit of a mess because we do have multiple workspaces that may wish to sign things at once. Closing this for now.

The best path forwards is probably to sign most things with a secret file key, and use the smartcard in only one workspace, at which point the gpg agent agent can be tied to that one workspace.

PR #804, opened 2018-03-29
Update selinux support

glevand:
Depends on https://github.com/coreos/portage-stable/pull/654 (Update selinux support) and https://github.com/coreos/coreos-overlay/pull/3155 (Update selinux support).


glevand:
Rebased to latest.

build_library/build_image_util.sh
```diff
@@ -506,6 +506,16 @@ EOF
         "${BUILD_DIR}/${image_kconfig}"
   fi
 
+  # Build the selinux policy and apply file labels.
+  if pkg_use_enabled coreos-base/coreos selinux; then
+    setup_qemu_static "${root_fs_dir}"
+    sudo chroot "${root_fs_dir}" /bin/bash -x << 'EOF'
+      (cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp)
+      setfiles -F /usr/lib/selinux/mcs/contexts/files/file_contexts /usr/lib/modules
```
dm0-: What is special about the kernel modules that require relabeling?

setup_board
```diff
@@ -265,6 +265,9 @@ PORTAGE_BINHOST="${BOARD_BINHOST}"
 # You can use --select to override this.
 EMERGE_DEFAULT_OPTS="--oneshot"
 
+# SELinux installed policies.
+POLICY_TYPES="mcs"
```
dm0-: Is there a reason to keep `POLICY_TYPES` in the base profile set to `targeted mcs mls`?  Maybe we should make this change there instead.  Or, if we don't want it in the base policy, maybe we can set this value in `profiles/coreos/targets/generic/make.defaults` so that there is only one file with this change when using multiple boards.

glevand: I missed that entry in the base profile.  I think `profiles/coreos/targets/generic/make.defaults` is the right place to set it since we are only building the mcs policy, and that is the same for all boards.  I'll move this change there.

build_library/build_image_util.sh
glevand: We build up the board's rootfs '${root_fs_dir}' from the SDK host, so the rootfs files don't have any selinux labels. The kernel module loader (modprobe, etc.) checks for proper labels, so without the relabeling here module loading will fail when selinux is enforcing.  I didn't look into why module loading worked with the old selinux policy/tools.

glevand:
Moved POLICY_TYPES change to base policy in https://github.com/coreos/coreos-overlay/pull/3155.

glevand:
@dm0- I sort of mentioned this on Slack before.  We should be labeling everything created ( I have a patch for it `Add SEL labels to all files` in https://github.com/glevand/coreos--scripts/commits/master), but there are still some things in the policy that need to be fixed up for Container Linux's startup.  As of now if we set all labels we get some SELinux `avc:  denied` messages at early startup.  It's something I planned to look into.  My current set of SELinux PRs will give slightly better SELinux support than what Container Linux has now.  I have a rebased set of patches ready.  I'll push out once testing finishes.

glevand:
Rebased to latest.

glevand:
Rebased to latest.

PR #808, opened 2018-04-18
image_to_vm: Cleanup memory settings

glevand:
Fix up inconsistent use of FLAGS_mem and '_get_vm_opt MEM', add function get_vm_mem to get more memory for arm64,  and remove unused files in the lib directory.


glevand:
Rebased to latest.

PR #810, opened 2018-04-23
transfer.sh: use update-percent in updateservicectl calls

sdemos:
cc: @bgilbert @ajeddeloh

### DIFF
--- a/offline_signing/transfer.sh
+++ b/offline_signing/transfer.sh
@@ -99,7 +99,8 @@ roll() {
     local board="$3"
 
     # Only ramp rollouts on AMD64; ARM64 is too small
-    if [[ "$board" = "amd64-usr" ]]; then
+    if [[ "$board" = "arm64-usr" ]]; then
+        echo "Setting rollout for arm64-usr to 100%"
         updateservicectl \
             --server="https://public.update.core-os.net" \
             --user="${ROLLER_USERNAME}" \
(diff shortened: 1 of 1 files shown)


offline_signing/transfer.sh
```diff
@@ -99,16 +99,19 @@ roll() {
...
             --key="${ROLLER_API_KEY}" \
             group update \
             --app-id="${APPID[${board}]}" \
             --group-id="${channel}" \
-            --update-count=3 \
-            --update-interval=60
+            --update-percent=100
+    else
+        # TODO(sdemos): update this behavior when rollout strategies land in roller
+        echo "Rollout set to 0%"
```
dm0-: The code looks like the update percent value is `-1` when unspecified, so will this actually leave it at its previous rate (100%) for amd64?

If it is set to zero, that's still a change in behavior.  Previously the `roll` subcommand would start rolling at the minimal rate, but now it stops updates from rolling.  Shouldn't it use 5% or something like that?  Otherwise it seems like there's not much reason to use the `roll` command (since arm64 is always at 100% anyway).

sdemos: My mistake, it definitely does just keep the old percentage. I'll add another call to set it explicitly. 

We can set it to a minimal percentage if we want. My thinking was we set everything to zero and then go from there. For this release, whoever is rolling it will have a script that will just bump the percentage by some amount and then sleep, so I was thinking they just run that immediately after this command and it does basically the same thing. I'm not particularly attached to that behavior though.

dm0-: Okay, I'm fine with it using a zero rate here if something else is managing the rate (although that sort of makes the `roll` command sound wrongly named).

sdemos: I agree that the name sounds wrong given what it's doing right now, but at least until we figure out exactly what the longer term plan is, I think we should just keep it named as is. Unless we decide to put it elsewhere, my current plan is to set an initial rollout strategy here, so once that lands, it should be right again.

sdemos:
ping

PR #812, opened 2018-05-03
create-basic-configdrive: update etcd v2 to etcd v3

mdjarv:
etcd v2 has been deprecated and is replaced with etcd v3 (etcd-member)

### DIFF
--- a/contrib/create-basic-configdrive
+++ b/contrib/create-basic-configdrive
@@ -26,7 +26,7 @@ This tool creates a basic config-drive ISO image.
 CLOUD_CONFIG="#cloud-config
 
 coreos:
-  etcd2:
+  etcd-member:
     name: <ETCD_NAME>
     advertise-client-urls: <ETCD_ADDR>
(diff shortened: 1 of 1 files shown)


bgilbert:
Thanks for the PR!

This script is pretty stale.  We should drop the fleet service while we're here.  Ideally the cloud-config would also be converted to Ignition, but I don't think that's strictly necessary.

mdjarv:
Good point, I didn't notice the fleet error until after I sent the PR

PR has been updated with the fleet removal but I did not look into the Ignition part

contrib/create-basic-configdrive
```diff
@@ -26,17 +26,15 @@ This tool creates a basic config-drive ISO image.
 CLOUD_CONFIG="#cloud-config
 
 coreos:
-  etcd2:
+  etcd-member:
```
dm0-: Does this change work?  It is not listed: https://github.com/coreos/coreos-cloudinit/blob/master/config/config.go#L39

PR #818, opened 2018-05-23
[WIP] build_torcx_store: support squashfs for ondisk addons

euank:
This can't be merged until the torcx binary in the image includes [squashfs support](https://github.com/coreos/torcx/pull/96).

This also is still WIP and may have some caveats.

The basic approach is to create both tgz and squashfs archives for anything included in the image, and just tgz for the ones that are only used by tectonic-torcx.
The build_image bit no longer includes the tgz files at all.

Everything else is juggling around files and code to try and make the above two statements true in a reasonable way.

### DIFF
--- a/build_image
+++ b/build_image
@@ -32,7 +32,7 @@ DEFINE_string base_pkg "coreos-base/coreos" \
   "The base portage package to base the build off of (only applies to prod images)"
 DEFINE_string base_dev_pkg "coreos-base/coreos-dev" \
   "The base portage package to base the build off of (only applies to dev images)"
-DEFINE_string torcx_manifest "${DEFAULT_BUILD_ROOT}/torcx/${DEFAULT_BOARD}/latest/torcx_manifest.json" \
+DEFINE_string torcx_manifest "${DEFAULT_BUILD_ROOT}/torcx/${DEFAULT_BOARD}/latest/torcx_ondisk_manifest.json" \
   "The torcx manifest describing torcx packages for this image (or blank for none)"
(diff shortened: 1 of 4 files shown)


euank:
Someone else should pick this up. Feel free to rebase / rewrite without too much regard for authorship if that's more convenient, this code was written at work and so is owned by Red Hat Inc. all the same.

This code, as is, has not been tested and almost certainly has issues.

The intended flow is that in the 'packages' step, `build_torcx_store` builds both squashfs + tgz from the same pkg root, and also creates a manifest for each (one with `path`+squashfs values for use in jenkins image step, the other with `url`+tgz values for tectonic torcx).

The images step will then download the new ondisk manifest to find the squashfs packages, while a human will still download the old manifest for signing and upload it to the final location.

The ondisk manifest does not need to be published anywhere, it's just a way for `packages` to communicate the list of squashfs files that the `images` step should inject (which the previous manifest was already doing via 'path').

I skimmed tectonic-torcx's code to make sure that it would probably behave correctly if no 'path' values were present.
I think that 'path' values with 'squashfs' paths may behave incorrectly in tectonic-torcx, but even if it did I don't think it can just be swapped out there since the 'hash' value is for all 'locations'.

I apologize for not having time to finish this up or leave more details.
Do note, if it seems reasonable to use a different bucket-management mechanism or Jenkins-communication mechanism, that could be worth doing instead and dropping much of this work.

cc @lucab

win-t:
Hi, any update on this?

euank:
Container Linux is deprecated. Closing this PR as stale.

PR #820, opened 2018-05-24
Remove dead code

ajeddeloh:
testing in jenkins


ajeddeloh:
passed jenkins, good to review

ajeddeloh:
Updated to switch to `--bind-gpg-agent=false`. We'll need to wait until the stable branch switches to this before removing the `--experimental` flag from `cork`

PR #821, opened 2018-05-31
Add signing server support to signing scripts

csssuf:
Any feedback before the test-run next week would be welcome.

Still TODO:
* ~Modify fero/these scripts to use detached signatures for users' signatures.~
* ~Torcx manifest signatures~


core_sign_update
```diff
@@ -18,9 +18,12 @@ export GCLIENT_ROOT=$(readlink -f "${SCRIPT_ROOT}/../../")
 DEFINE_string image "" "The filesystem image of /usr"
 DEFINE_string kernel "" "The kernel image"
 DEFINE_string output "" "Output file"
-DEFINE_string private_keys "" "Path or pkcs11 URI to private keys."
+DEFINE_string private_keys "" "Path, pkcs11 URI, or fero:<keyname> for private keys."
 DEFINE_string public_keys "" "Path to public keys in .pem format."
 DEFINE_string keys_separator ":" "Separator for the above keys"
+DEFINE_string user_signatures "" "Paths to user signatures to provide to signing server"
```
ajeddeloh: it looks like therse are colon seperated. That should be included in the description

core_sign_update
```diff
@@ -63,6 +66,14 @@ for key in "${private_keys[@]}"; do
 done
 signature_sizes="${signature_sizes:1:${#signature_sizes}}"
 
+IFS=":" read -a user_signatures <<< "$FLAGS_user_signatures"
```
ajeddeloh: a comment explaining the bash manipulation going on would be useful

core_dev_sign_update
```diff
@@ -0,0 +1,48 @@
...
+eval set -- "${FLAGS_ARGV}"
+
+cleanup() {
+	rm -f update
+	rm -f update.hash
+}
+
+trap cleanup INT TERM EXIT
+
+signature_sizes=""
```
ajeddeloh: a comment explaining what this is doing would be nice

core_sign_update
csssuf: I used `read` here to match the way `$FLAGS_private_keys`/`$FLAGS_public_keys` are handled elsewhere in this script, but I don't have a preference between the two.

ajeddeloh:
Feedback from the test run:
 - Would be nice to have a comprehensive wrapper signing script that wrapped everything so you could just say "sign the directory you got with transfer.sh"
 - we should standardize the locations signatures are generated in / are consumed in
 - Needs a PR to release checklists with the new signing steps
 - Script output could be more clear about success / failure (not that our current ones do this)
 - Evenutally the release signing bit itself could be moved to jenkins (out of scope here)

ajeddeloh:
Yeah, we could even keep a git repo of sigs, so devs signing could just use `--output-dir /path/to/git/repo` and then push their sigs up and the releaser can pull them down

ajeddeloh:
Ran through this again. New scripts are much friendlier. We used git and that worked great. I still want a wrapper than you can just say `dev_sign -a xxxxx -b xxxxxx -s xxxxxx` but that can come later

csssuf:
I think this should be ready to merge (other than potentially needing to change the default fero IP in `signing/sign.sh` once it's actually deployed). Any concerns, @ajeddeloh / @bgilbert ?

bgilbert:
SGTM.

PR #822, opened 2018-06-15
Add check for gpg tool

Akasurde:
if gpg tool is not available then we should should break early.

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>


contrib/create-coreos-vdi
```diff
@@ -39,6 +39,13 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+# GPG tools required
+which gpg &>/dev/null
+if [ $? -ne 0 ]; then
+    echo "$0: GPG tool is required to verify GPG signature." >&2
```
lucab: Nit: s/GPG tool/gpg/

Akasurde:
@lucab Done.

contrib/create-coreos-vdi
```diff
@@ -39,6 +39,13 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+# GPG tools required
+which gpg &>/dev/null
+if [ $? -ne 0 ]; then
```
webknjaz: In general, you'd want to wrap vars with double quotes `"$?"`. It's just that it this case it would always have some return code, but for arbitrary var it might end up being interpolated as `if [  -ne 0 ]; then` in case of empty var/string, which would lead to syntax error being raised by shell interpreter.

PR #824, opened 2018-06-19
transfer.sh: changes to updateservicectl usage for roller rollout strategies

sdemos:
This pr changes the update percent setting to use the new `group percent` `updateservicectl` command, and creates a new `transfer.sh` command call `roll` (the old `roll` is renamed to `ready`) which creates a linear rollout that reaches 100% in the provided number of seconds for each channel. `roll` only sets up a rollout for the amd64 application (the arm64 application is just set to 100% by default, so there is no need for a rollout).

 The old `roll` command, now called `ready`, sets the update percent for amd64 to 0% (and arm64 to 100%) and updates the channel to point at the provided package version. 

This is the only place in the scripts repo that needs to be updated for the new version of updateservicectl. It relies on the new version of updateservicectl to be in the sdk, and also relies on the new version of coreupdate to be deployed (it's not yet). 

cc: @bgilbert @dm0- @ajeddeloh

### DIFF
--- a/offline_signing/transfer.sh
+++ b/offline_signing/transfer.sh
@@ -93,33 +93,22 @@ upload() {
     fi
 }
 
-roll() {
+ready() {
     local channel="$1"
     local version="$2"
     local board="$3"
 
-    # Only ramp rollouts on AMD64; ARM64 is too small
-    if [[ "$board" = "arm64-usr" ]]; then
-        echo "Setting rollout for arm64-usr to 100%"
-        updateservicectl \
-            --server="https://public.update.core-os.net" \
-            --user="${ROLLER_USERNAME}" \
-            --key="${ROLLER_API_KEY}" \
-            group percent \
-            --app-id="${APPID[${board}]}" \
-            --group-id="${channel}" \
-            --update-percent=100
-    else
-        echo "Rollout set to 0%"
-        updateservicectl \
-            --server="https://public.update.core-os.net" \
-            --user="${ROLLER_USERNAME}" \
-            --key="${ROLLER_API_KEY}" \
-            group percent \
-            --app-id="${APPID[${board}]}" \
-            --group-id="${channel}" \
-            --update-percent=0
-    fi
+    # setting the percent will deactivate (not delete) any existing rollouts for
+    # this specific group.
+    echo "Rollout set to 0% for ${board}"
+    updateservicectl \
+        --server="https://public.update.core-os.net" \
+        --user="${ROLLER_USERNAME}" \
+        --key="${ROLLER_API_KEY}" \
+        group percent \
+        --app-id="${APPID[${board}]}" \
+        --group-id="${channel}" \
+        --update-percent=0
 
     # FIXME(bgilbert): We set --publish=true because there's no way to
     # say --publish=unchanged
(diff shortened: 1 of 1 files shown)


ajeddeloh:
You'll want to PR the release checklists too with the new release instructions

offline_signing/transfer.sh
```diff
@@ -93,7 +93,7 @@ upload() {
     fi
 }
 
-roll() {
+ready() {
```
ajeddeloh: This should probably also set arm to 0% and the new roll can set it to 100%, right?

offline_signing/transfer.sh
```diff
@@ -135,9 +134,26 @@ roll() {
...
+        --group-id="${channel}" \
+        --duration="${seconds}" \
+        --frame-size="60"
+}
+
 usage() {
     echo "Usage: $0 {download|upload} <ARTIFACT-DIR> [{-a|-b|-s} <VERSION>]..." >&2
-    echo "Usage: $0 roll [{-a|-b|-s} <VERSION>]..." >&2
+    echo "Usage: $0 ready [{-a|-b|-s} <VERSION>]..." >&2
+    echo "Usage: $0 roll [{-a|-b|-s} <SECONDS-TO-100-PERCENT>]..." >&2
```
bgilbert: Please don't make me type `86400`.  :cry:

offline_signing/transfer.sh
sdemos: yeah, we can do it that way if you want. I just left the old `roll` command as is, and it just sets the update percent to 100%.

offline_signing/transfer.sh
sdemos: would you rather have it as hours? days? I'm open for whatever here.

offline_signing/transfer.sh
ajeddeloh: I think we do want to do that. It should be doing roughly the same thing for both amd64 and arm64, It's just that "slowly roll out" and "roll out all at once" are equivilent for arm. We shouldn't have a command prepare to rollout on amd and do the rollout on arm

offline_signing/transfer.sh
ajeddeloh: If it comes to that, we can use the raw updateservicectl commands too.

sdemos: or just modify the script sometime later. I'll set it to hours.

@arithx I agree in general that allowing arbitrary duration units would be ideal, but that's not how I ended up writing it for `updateservicectl` and I have no clue how to do that in bash (and I'm not sure I want to know)

sdemos:
@bgilbert @ajeddeloh fixed both of your issues.

offline_signing/transfer.sh
```diff
@@ -93,11 +93,41 @@ upload() {
...
     local channel="$1"
     local version="$2"
     local board="$3"
 
+    echo "Rollout set to 0% for ${board}"
+    updateservicectl \
+        --server="https://public.update.core-os.net" \
+        --user="${ROLLER_USERNAME}" \
+        --key="${ROLLER_API_KEY}" \
+        group percent \
```
bgilbert: Will this terminate any existing rollout in the group?

offline_signing/transfer.sh
```diff
@@ -177,26 +197,55 @@ case "${CMD}" in
...
+        # Walk argument pairs.
+        while [[ $# > 0 ]]; do
+            c="$1"
+            s="${2?Must provide a time in hours}"
+            shift 2
+
+            case "${c}" in
+                -a)
+                    $CMD "alpha" "${s}" "amd64-usr"
+                    $CMD "alpha" "${v}" "arm64-usr"
```
bgilbert: `${v}` is not defined.

Since the only difference between `roll` and the other subcommands is the error message for `$2`, I'd suggest making only that one assignment conditional, or making nothing conditional and using a more generic error message.

offline_signing/transfer.sh
sdemos: yeah, part of the behavior of setting a percent anywhere is that it will pause the existing rollout. 

it's less termination and more pausing, it would be possible to unpause it (although in this specific case it doesn't make much sense). but, for instance, if someone changes the percent in the web ui at some point, that would also pause the automated rollout.

offline_signing/transfer.sh
sdemos: I can make it a more generic error message if `$2` is not assigned. that makes more sense than my current approach.

sdemos:
@bgilbert fixed your notes

offline_signing/transfer.sh
bgilbert: Probably also needs a comment to that effect.

offline_signing/transfer.sh
```diff
@@ -180,23 +200,23 @@ esac
...
-        $CMD "beta" "${v}" "amd64-usr"
-        $CMD "beta" "${v}" "arm64-usr"
-        ;;
-    -s)
-        $CMD "stable" "${v}" "amd64-usr"
-        ;;
-    *)
-        usage
-        ;;
+        -a)
```
bgilbert: Please don't mix in random indentation changes.  The indentation that was there is traditional for bash `case` statements.

sdemos: sorry, that was unintended, my editor just did that for me and the rebasing made it slip through the cracks.

PR #831, opened 2018-07-23
DNM Rm dev image

ajeddeloh:


ajeddeloh:
Ready for review.

ajeddeloh:
Fixed.

build_image
```diff
@@ -146,8 +143,6 @@ fi
 mkdir -p "${BUILD_DIR}"
 
 DISK_LAYOUT="${FLAGS_disk_layout:-base}"
-DEV_DISK_LAYOUT="${FLAGS_disk_layout:-devel}"
```
dm0-: I think this is the only place the `devel` layout is used, so that could be dropped as well.

ajeddeloh: Done (and verified the dev container and prod image still build).

PR #846, opened 2018-10-18
Drop some useless paths on prod images

dm0-:


dm0-:
They shouldn't be using `/usr/include` for anything.  It didn't exist in the image until `json-c` was added a couple releases again, and right now it just contains a symlink loop.

PR #848, opened 2018-10-25
signing/sign.sh: fix armoring

ajeddeloh:
We use gpg --enarmor to convert the torcx manifest binary sigs to ascii
armored ones. Unfortunately --enarmor just wraps the binary blob without
realizing its a signature which breaks torcx. Pipe the output through
sed to fix the header.

### DIFF
--- a/signing/sign.sh
+++ b/signing/sign.sh
@@ -56,7 +56,10 @@ fero-client \
 	--secret-key coreos-app-signing-key \
 	${torcx_signature_arg}
 gpg2 --enarmor \
-	--output "${DATA_DIR}/torcx_manifest.json.asc" \
-	"${DATA_DIR}/torcx_manifest.json.sig-fero"
+	--output - \
+	"${DATA_DIR}/torcx_manifest.json.sig-fero" \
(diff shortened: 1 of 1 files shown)


signing/sign.sh
```diff
@@ -56,7 +56,10 @@ fero-client \
 	--secret-key coreos-app-signing-key \
 	${torcx_signature_arg}
 gpg2 --enarmor \
-	--output "${DATA_DIR}/torcx_manifest.json.asc" \
-	"${DATA_DIR}/torcx_manifest.json.sig-fero"
+	--output - \
+	"${DATA_DIR}/torcx_manifest.json.sig-fero" \
+	| sed 's/ARMORED FILE/SIGNATURE/' \
```
bgilbert: Ridiculous nit: `ARMORED FILE` could theoretically show up in the signature body, but `-` can't, so probably better to check for the trailing `-----`.

ajeddeloh: Are spaces allowed? I don't think they are.

bgilbert: Ah, you're right.

PR #857, opened 2018-11-16
signing/sign.sh: update fero ip

ajeddeloh:


ajeddeloh:
I believe so. I think @pbx0 set that up. @pbx0 can you confirm?

ajeddeloh:
Merging regardless, since that's where it's at now.

PR #870, opened 2019-05-31
jenkins: Free more disk space on successful job completion

dm0-:
This should make it safer for scheduling multiple runs on small workers without running out of disk space due to multiple workspaces having duplicate SDKs.


dm0-:
This moves the actual disk cleanup to coreos/jenkins-os#88.  It moves ccache files outside the deleted directories to try to improve build time in repeated runs, but that might need to be dropped if it gets too big.

PR #876, opened 2019-08-23
setup_board: add workaround for binutils issue

ajeddeloh:
Add a workaround to be sure we're using the correct binutils when SDK
sharing.

### DIFF
--- a/setup_board
+++ b/setup_board
@@ -339,6 +339,11 @@ if [ $FLAGS_default -eq $FLAGS_TRUE ] ; then
   echo $BOARD_VARIANT > "$GCLIENT_ROOT/src/scripts/.default_board"
 fi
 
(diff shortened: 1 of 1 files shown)


setup_board
```diff
@@ -339,6 +339,11 @@ if [ $FLAGS_default -eq $FLAGS_TRUE ] ; then
   echo $BOARD_VARIANT > "$GCLIENT_ROOT/src/scripts/.default_board"
 fi
 
+# workaround for https://wiki.gentoo.org/wiki/Binutils_2.32_upgrade_notes/elfutils_0.175:_unable_to_initialize_decompress_status_for_section_.debug_info
+# can be dropped once all channels use 2.32.
```
bgilbert: Let's explicitly list the CL version that must reach stable before this workaround can be dropped.

ajeddeloh: done

PR #885, opened 2019-12-13
build_library/disk_util: work around losetup bug

ajeddeloh:
Retry losetup if it fails, up to 5 times with 5 seconds between retries.


ajeddeloh:
https://jenkins-container-linux.apps.ci.centos.org/job/os/job/manifest/463/cldsv/

Will backport to other branches once it passes CI and some nightly since it doesn't affect the images and will help make the update-linux job more useful

PR #888, opened 2019-12-16
disk_util: retry loopback mounts if they fail

ajeddeloh:
This fails frequently but should succeed if retried. This should reduce
failed builds.

### DIFF
--- a/build_library/disk_util
+++ b/build_library/disk_util
@@ -598,9 +598,22 @@ def Mount(options):
       mount_opts.append('subvol=%s' % mount['fs_subvolume'])
 
     Sudo(['mkdir', '-p', full_path])
-    Sudo(['mount', '-t', mount.get('fs_type', 'auto'),
-                   '-o', ','.join(mount_opts),
-                   options.disk_image, full_path])
+    # This tends to fail, retry if it does
+    err = None
+    for i in range(0,5):
+      try:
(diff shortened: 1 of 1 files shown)


ajeddeloh:
https://jenkins-container-linux.apps.ci.centos.org/job/os/job/manifest/472/cldsv/

build_library/disk_util
```diff
@@ -598,9 +598,22 @@ def Mount(options):
...
+                       '-o', ','.join(mount_opts),
+                       options.disk_image, full_path])
+        err = None
+        break
+      except subprocess.CalledProcessError as e:
+        print("Error mounting %s, attempt %d" % (full_path, i))
+        err = e
+        time.sleep(5)
+
+    if err is not None:
```
jlebon: One funky but cool Python construct you can use here is the `for-else`:

```python
for ...:
  ...
else:
  raise err
```

It'll only raise if we didn't explicitly break from the loop. That means you can also drop the `err = None` here. Pretty sure you can also drop the initial `err = None` too since Python doesn't scope variables in nested control flow blocks.

ajeddeloh: Maybe a follow up PR that does that here and where `disk_util` runs `losetup` explicitly? I want to get this backported asap so the update-linux jobs can actually give good results.

jlebon: More of a :stars: The More You Know :stars:  thing. Definitely not a blocker!

